PR #20, opened 2024-07-30
ci: add lint step for pr workflow

starbops:


starbops:
After #21 merged, we can rebase the PR.

Issue #2, opened 2024-02-20
[FEATURE] Redfish support


starbops:
**Is your feature request related to a problem? Please describe.**

Currently, KubeBMC only supports limited IPMI functionalities for KubeVirt VMs. It would also be helpful to support Redfish as it is newer, feature-rich, and gradually becomes dominant in such areas.

**Describe the solution you'd like**

A set of Redfish APIs that supports users to communicate with to instruct KubeBMC to:

- Retrieve power status,
- Set power on/off/reset,
- Set boot device, and
- Mount/unmount virtual media

for KubeVirt VMs.

**Describe alternatives you've considered**

N/A

**Additional context**

We could implement the following resources to provide a minimum viable functionalities. Kudos to @rwleea!

```json
{
    "@odata.type": "#ServiceRoot.v1_16_1.ServiceRoot",
    "@odata.id": "/redfish/v1/",
    "Id": "RootService",
    "Name": "Root Service",
    "Vendor": "Contoso",
    "Product": "WidgetDeluxe 8744",
    "RedfishVersion": "1.6.0",
    "UUID": "92384634-2938-2342-8820-489239905423",
    "Systems": {
        "@odata.id": "/redfish/v1/Systems"
    }
}
```

```json
{
  "@odata.type": "#ComputerSystemCollection.ComputerSystemCollection",
  "Name": "Computer System Collection",
  "@odata.id": "/redfish/v1/Systems",
  "Members@odata.count": 1,
  "Members": [
    {
      "@odata.id": "/redfish/v1/Systems/1"
    }
  ]
}
```

```json
{
  "@odata.id": "/redfish/v1/Systems/1",
  "@odata.type": "#ComputerSystem.v1_22_0.ComputerSystem",
  "Id": "1",
  "Name": "My Computer System",
  "UUID": "00000000-0000-0000-0000-000000000000",
  "HostName": "web-srv344",
  "Status": {
    "State": "Enabled",
    "Health": "Warning",
    "HealthRollup": "Warning",
    
  },
  "PowerState": "On",
  "Boot": {
    "BootSourceOverrideEnabled": "Once",
    "BootSourceOverrideMode": "UEFI",
    "BootSourceOverrideTarget": "Pxe",
    "BootSourceOverrideTarget@Redfish.AllowableValues": [
      "None",
      "Pxe",
      "Cd",
      "Hdd"
    ]
  },
  "VirtualMedia": {
[...]

PR #39, opened 2024-11-20
feat: add redfish support

starbops:
Support basic (a limited set) Redfish resource retrieval and actions including:

- get power status
- power actions
- set boot device

Redfish resource models and API services are generated from the Redfish OpenAPI spec (~~trimmed and based on [DSP8010 2024.03](https://www.dmtf.org/dsp/DSP8010)~~ based on [DSP8010 2023.03](https://www.dmtf.org/dsp/DSP8010)) via [openapi-generator](https://github.com/OpenAPITools/openapi-generator).

Related issue: #2


starbops:
For better compatibility, I decided to downgrade the Redfish schema a bit from 2024.3 to 2023.3.

PR #98, opened 2025-11-07
fix(helm): remove duplicate keys in ClusterRoleBinding

mcanevet:
The metadata.labels and metadata.name keys are duplicate in the ClusterRoleBinding.

PR #99, opened 2025-11-09
ci: pin yamale version to 6.0.0 for ct ci

starbops:
This PR tries to fix up recent chart CI workflow failures, e.g., https://github.com/starbops/kubevirtbmc/actions/runs/19167891423/job/54864954127


starbops:
The fix will be included in #98. Close this one as duplicate.

Issue #83, opened 2025-06-10
[ENHANCEMENT] Create v1beta1 API design


0xFelix:
**Is your enhancement request of an existing feature related to a problem? Please describe.**

With the lessons learned from the v1alpha1 API a new v1beta1 API needs to be designed.

**Describe the solution you'd like**

v1beta1 API with the following changes compared to v1alpha1

- No cross namespacing of VMs (i.e. BMCs must be created in the same namespace as the VM they are interfacing with)
- Credentials should be referenced from `Secret`
- Rename `ServiceIP` to `ClusterIP`

**Describe alternatives you've considered**

**Additional context**

This is in preparation to move this repo to the KubeVirt org.

starbops:
@salawhaaat feel free to assign yourself to the ticket.

salawhaaat:
> [@salawhaaat](https://github.com/salawhaaat) feel free to assign yourself to the ticket.

To be able to assign issues to myself or to others I have to be the owner or a collaborator of the repository. Can you make collaborator? 

Thank you @starbops !

starbops:
We need to work on this again carefully (both the API and the corresponding controllers). Most of the following features/fixes will depend on this.

starbops:
According to the issue description, we need to address at least the following concerns in the new v1beta1 API and controller implementation:

- [ ] #62 
- [ ] #4 
- [ ] Rename `ServiceIP` to `ClusterIP`

We don't consider providing backward compatibility to the v1alpha1 API. Besides, the API group will be changed from `virtualmachine.kubevirt.io` to `bmc.kubevirt.io`.

Once the v1beta1 API and its controller implementation are finalized, the following items can develop from it (not included in the scope of this issue):

- [ ] #44
- [ ] #3
- [ ] #47

PR #110, opened 2025-12-01
feat: refractor the api design to v1beta1 and respective controller loop as well

anishbista60:
Follow up issue: #83 

- Introduced the new api v1beta1
- Drop the logic of the CR creation by controller. 
-  Added the validation logic when CR is created in wrong ns where VM is not there (Updated the condition of CR and Log the not found in controller manager)
- Added the rolebinding the sa creation in the reconcile loop and set them as the owner of the kubevirtbmc CR
- Modified the test case for the controller. 
- Modified the e2e test

### DIFF
--- a/api/bmc/v1beta1/virtualmachinebmc_types.go
+++ b/api/bmc/v1beta1/virtualmachinebmc_types.go
@@ -46,7 +46,7 @@ type VirtualMachineBMCStatus struct {
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
-// +kubebuilder:resource:shortName=vmbmc;vmbcms
+// +kubebuilder:resource:shortName=vmbmc;vmbmcs
 
 // VirtualMachineBMC is the Schema for the virtualmachinebmcs API
 type VirtualMachineBMC struct {
--- a/config/crd/bases/bmc.kubevirt.io_virtualmachinebmcs.yaml
+++ b/config/crd/bases/bmc.kubevirt.io_virtualmachinebmcs.yaml
@@ -13,7 +13,7 @@ spec:
     plural: virtualmachinebmcs
     shortNames:
     - vmbmc
-    - vmbcms
+    - vmbmcs
     singular: virtualmachinebmc
   scope: Namespaced
   versions:


anishbista60:
To accomplish https://github.com/starbops/kubevirtbmc/issues/62, we need to think about below 
 
So, right now service account is created the `kubevirbmc-system` that is used by agent pod (https://github.com/starbops/kubevirtbmc/blob/main/config/virtbmc/service_account.yaml#L12). Creating the BMC CR and correponding pod and services at same ns as VM would require serviceaccount in that ns where VM is deployed. Let's say there are 10 VM across 10 ns. How we handle present of service account, role, role binding within that ns?

starbops:
We can handle the CI issue in other PRs. Let's focus on the API and controller now.

anishbista60:
> * AFAIK, we need to drop clusterole and binding and replace with role and role binding with sa in for each ns where vm is deployed and this part should be done by user.
> *

Not role and rolebinding but only `sa`. May leverage cluster role and cluster role binding which is a single time effort during `make deploy` (but it will cluster wide). Let's say there are 10 different ns and then sa will be the same across all the ns which will be using cluster  role and binding.

cmd/main.go
starbops: Is this file really required?

internal/controller/service/controller_test.go
```diff
@@ -55,25 +55,6 @@ var _ = Describe("Service Controller", func() {
...
-				},
-				Spec: virtualmachinev1.VirtualMachineBMCSpec{
-					Username:                testUsername,
-					Password:                testPassword,
-					VirtualMachineNamespace: testVMNamespace,
-					VirtualMachineName:      testVMName,
-				},
-			}
-			Expect(k8sClient.Create(ctx, virtualMachineBMC)).To(Succeed())
-
```
starbops: Why is this removed? How does the VirtualMachineBMC CR get updated without it being created first?

internal/controller/virtualmachinebmc/controller.go
```diff
@@ -131,11 +201,55 @@ func (r *VirtualMachineBMCReconciler) constructServiceFromVirtualMachineBMC(virt
...
+				{
+					Type:   virtualmachinev1.ConditionReady,
+					Status: metav1.ConditionFalse,
+					Reason: "VirtualMachineNotFound",
+					Message: fmt.Sprintf("VirtualMachine %q not found in namespace %q",
+						virtualMachineBMC.Spec.VirtualMachineRef.Name,
+						virtualMachineBMC.Namespace),
+					LastTransitionTime: metav1.Now(),
+				},
+			}
```
starbops: The condition update part appears in different places; it'd be nice if we could make it a utility function. For instance:

```
virtualMachineBMC.SetCondition(bmcv1.ConditionReady, metav1.ConditionFalse, "VirtualMachineNotFound", fmt.Sprintf(...))
```

starbops: We can use the [`SetStatusCondition`](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/meta#SetStatusCondition) to reduce the potential excessive status update calls:

```suggestion

			if changed := meta.SetStatusCondition(&cronJob.Status.Conditions, metav1.Condition{
				Type:    bmcv1.ConditionReady,
				Status:  metav1.ConditionFalse,
				Reason:  "VirtualMachineNotFound",
				Message: fmt.Sprintf("VirtualMachine %q not found in namespace %q",
						virtualMachineBMC.Spec.VirtualMachineRef.Name,
						virtualMachineBMC.Namespace),
			}); changed {
				return false, r.Status().Update(ctx, virtualMachineBMC)
			}
			return false, nil
```

starbops (commented):
First round review. Please see below.

starbops:
> When somebody delete the `sa` and `rolebinding`. It will not reconcile it as per the latest commit . I will add that part as well later today

Yeah, need to trigger the VirtualMachineBMC reconciliation. `Owns()` can do the trick. Also, better to set the owner reference to those RBAC resources.

cmd/main.go
anishbista60: yes, it was required to generate the new api but i think we can drop now.

internal/controller/service/controller.go
```diff
@@ -77,8 +78,17 @@ func (s *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-	virtualMachineBMC.Status.ServiceIP = svc.Spec.ClusterIP
+	virtualMachineBMC.Status.Conditions = []metav1.Condition{
+		{
+			Type:               bmcv1.ConditionReady,
+			Status:             metav1.ConditionTrue,
+			Reason:             "ServiceReady",
+			Message:            "ClusterIP assigned to the Service",
+			LastTransitionTime: metav1.Now(),
+		},
+	}
```
starbops: Same, we can use `SetStatusCondition` here.

internal/controller/virtualmachinebmc/controller.go
```diff
@@ -227,9 +357,79 @@ func (r *VirtualMachineBMCReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-		For(&virtualmachinev1.VirtualMachineBMC{}).
+		For(&bmcv1.VirtualMachineBMC{}).
 		Owns(&corev1.Pod{}).
 		Owns(&corev1.Service{}).
+		Owns(&corev1.ServiceAccount{}).
+		Owns(&rbacv1.RoleBinding{}).
+		Watches(
+			&kubevirtv1.VirtualMachine{},
+			handler.EnqueueRequestsFromMapFunc(r.findVirtualMachineBMCsForVM),
+		).
```
starbops: I'm sure this can benefit us someday, but could you elaborate on which part currently needs the VirtualMachineBMC reconciled upon VirtualMachine changes?

starbops:
@anishbista60 We can implement the webhook in another PR. I'll create an issue to track what we want to validate/mutate.

anishbista60:
> I believe we can drop the v1beta1 API completely, as it is no longer referenced. Please do so and update the `PROJECT` file accordingly.

Are you sure to drop v1beta1 😆 ? 
I know it's v1alpha1

starbops:
> > I believe we can drop the v1beta1 API completely, as it is no longer referenced. Please do so and update the `PROJECT` file accordingly.
> 
> Are you sure to drop v1beta1 😆 ? I know it's v1alpha1

LOL yes you're right! I need to grab a cup of coffee...

internal/controller/virtualmachinebmc/controller.go
```diff
@@ -42,22 +48,97 @@ type VirtualMachineBMCReconciler struct {
...
+			Kind:     "ClusterRole",
+			Name:     clusterRoleName,
+		},
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      "ServiceAccount",
+				Name:      serviceAccountName,
+				Namespace: virtualMachineBMC.Namespace,
+			},
+		},
```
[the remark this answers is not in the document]
anishbista60: Does it make sense to make `kind` as constant ? For name it is fine.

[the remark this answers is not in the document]
starbops: I can't find any usable constant for `ClusterRole` from `rbacv1`. You can keep the original change, it's fine.

api/bmc/v1beta1/virtualmachinebmc_types.go
```diff
@@ -14,49 +14,40 @@ See the License for the specific language governing permissions and
...
+	// List of current conditions (e.g., Ready)
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
 }
 
-//+kubebuilder:object:root=true
-//+kubebuilder:subresource:status
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:shortName=vmbmc
+// +kubebuilder:resource:shortName=vmbmcs
```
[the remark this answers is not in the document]
anishbista60: I was using comma and getting issue. Silly mistake.

PR #118, opened 2025-12-04
Feat: Leverage secret to read the credentials for agent pod

anishbista60:
Tracking issue: https://github.com/starbops/kubevirtbmc/issues/112, https://github.com/starbops/kubevirtbmc/issues/4, https://github.com/starbops/kubevirtbmc/issues/61,  https://github.com/starbops/kubevirtbmc/issues/43

- Populate  the Env section in the agent pod
- Created the reconcillation logic for the secret and setup the watches for secret to trigger the reconcile loop for  the bmc when ever any changes happen to the secret. 
- Delete the pod when there will be any changes to the secret field. 
- Updated the test file 
- Read the credential from the reference secret. 
- Add the basic auth support for redfish
- Donot create the pod when either secret or  vm does not exist. 

> [!NOTE]
> In this PR IPMI, authentication will not work cause we need to create the backend server to handle that part which is no present now.

### DIFF
--- a/pkg/session/token.go
+++ b/pkg/session/token.go
@@ -9,6 +9,12 @@ import (
 	"sync"
 )
 
+type contextKey string
+
+const (
+	userContextKey contextKey = "user"
+)
+
 var ts TokenStore
 
 type TokenStore struct {
@@ -86,25 +92,34 @@ func AuthMiddleware(bmcUser, bmcPassword string) func(next http.Handler) http.Ha
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			user, pass, ok := r.BasicAuth()
 
-			if !ok || user != bmcUser || pass != bmcPassword {
-				// Also check for X-Auth-Token for session-based authentication
-				token := r.Header.Get("X-Auth-Token")
-				if token != "" {
-					ts.rwMutex.RLock()
-					_, exists := ts.store[token]
-					ts.rwMutex.RUnlock()
-					if exists {
-						next.ServeHTTP(w, r)
-						return
-					}
+			if ok {
+				if user != bmcUser || pass != bmcPassword {
+					w.Header().Set("WWW-Authenticate", `Basic realm="Redfish"`)
+					http.Error(w, "Unauthorized", http.StatusUnauthorized)
+					return
 				}
+				ctx := context.WithValue(r.Context(), userContextKey, user)
+				r = r.WithContext(ctx)
+				next.ServeHTTP(w, r)
+				return
+			}
 
+			token := r.Header.Get("X-Auth-Token")
+			if token == "" {
 				w.Header().Set("WWW-Authenticate", `Basic realm="Redfish"`)
 				http.Error(w, "Unauthorized", http.StatusUnauthorized)
 				return
 			}
-			ctx := context.WithValue(r.Context(), "user", user)
-			r = r.WithContext(ctx)
+
+			ts.rwMutex.RLock()
+			_, exists := ts.store[token]
+			ts.rwMutex.RUnlock()
+
+			if !exists {
+				w.Header().Set("WWW-Authenticate", `Basic realm="Redfish"`)
+				http.Error(w, "Unauthorized", http.StatusUnauthorized)
+				return
+			}
 			next.ServeHTTP(w, r)
 		})
 	}


anishbista60:
lol, build failed. We need to fix this issue cause it is coming ocassionally due to the webhook failure. CR resource be applied after some time when manager is up and running.

anishbista60:
FYL: I have add delay of 10 sec. It is working now.

starbops:
> In this PR IPMI, authentication will not work cause we need to create the backend server to handle that part which is no present now.

Does that mean the IPMI authentication stay the same (with hardcoded default credentials), or entirely unusable?

internal/controller/virtualmachinebmc/controller.go
```diff
@@ -53,8 +53,9 @@ const (
 )
 
 var (
-	ownerKey = ".metadata.controller"
-	apiGVStr = bmcv1.GroupVersion.String()
+	ownerKey   = ".metadata.controller"
+	apiGVStr   = bmcv1.GroupVersion.String()
+	secretKeys = []string{"username", "password"}
```
starbops: We can make these two as constants.

pkg/virtbmc/virtbmc.go
```diff
@@ -22,6 +21,14 @@ type Options struct {
 	Address        string
 	IPMIPort       int
 	RedfishPort    int
+	SecretRef      string
```
starbops: Seems to be unused?

cmd/virtbmc/main.go
```diff
@@ -64,6 +64,9 @@ func main() {
 			},
 		},
 		Action: func(cCtx *cli.Context) error {
+			options.BMCUser = os.Getenv("BMC_USERNAME")
+			options.BMCPassword = os.Getenv("BMC_PASSWORD")
```
starbops: Can we add logic to bail out the execution if any environment variable is not provided? We do not allow a virtual BMC running with the default credentials.

anishbista60: The agent pod always receives the BMC credentials from a Kubernetes Secret.
These environment variables are injected by the controller, and if someone manually edits the pod and removes them, the reconcile loop will restore the correct configuration immediately.

Also, default Redfish credentials are no longer used; authentication only happens through the secret.

 AFAIK, Practically the “env not provided” scenario will never occur.
 
 Can you think about such situation where env will be missing ?

anishbista60: That said, adding a fail-fast check is still acceptable as a defensive measure but if there won't be any scenario where `env  is missing` then I think adding the check part will remain useless.

starbops: You're correct, but please think of it from a command-line tool perspective. The BMC username and password are essential to its execution. The command cannot function without them, just like missing the kubeconfig file. Neither do we provide logic to handle missing those values, nor do we fail execution immediately. How should the command work in such a circumstance?

I would suggest panicking and ending the execution immediately. This should be just a couple of lines of code. It's no harm to do it. We will inevitably have to face the issue when writing tests.

anishbista60: Thanks for input.

cmd/virtbmc/main.go
```diff
@@ -64,6 +65,13 @@ func main() {
...
+
+			if options.BMCUser == "" || options.BMCPassword == "" {
+				log.Panic("BMC credentials missing: both BMC_USERNAME and BMC_PASSWORD must be provided")
```
starbops: Please use `logrus.Panic()` or simply `panic()` instead of introducing new dependency.

internal/controller/virtualmachinebmc/controller.go
```diff
@@ -215,6 +245,31 @@ func (r *VirtualMachineBMCReconciler) constructServiceFromVirtualMachineBMC(virt
...
+		if apierrors.IsNotFound(err) {
+			log.V(1).Info("virtBMC Pod already absent", "pod", podName)
+			return nil
+		}
+
+		log.Error(err, "unable to delete virtBMC Pod", "pod", podName)
+		return err
+	}
+
+	log.Info("Deleted virtBMC Pod after Secret change", "pod", podName)
```
starbops: It feels irrelevant for this log to appear in the method. Maybe there is another better place. I suppose move it to the caller?

Issue #44, opened 2024-12-19
[FEATURE] Support virtual media mount


starbops:
**Is your feature request related to a problem? Please describe.**

This long-awaited feature could make several virtual machine provisioning use cases possible. With this feature implemented, we will be able to dynamically attach an ISO image to the target virtual machine and boot from it.

**Describe the solution you'd like**

My initial thought is, this will need a third-party storage solutions to handle the volume lifecycle. We should do our best to make our part as neutral as possible, at least giving users the ability to choose which solution to use via configuration. As I use Harvester a lot, [Longhorn](https://github.com/longhorn/longhorn) seems a good fit. However, [CDI](https://github.com/kubevirt/containerized-data-importer) could also be a valid candidate. With that said, both are a bit overkill for our use case. If this is infeasible or intolerable, we must handle the image and volume ourselves.

**Describe alternatives you've considered**

N/A

**Additional context**

This is one of the remaining item from #2.

starbops:
This is how people deal with virtual media mount with the sushy-tools BMC emulator, hope this can shed some lights:
https://docs.openstack.org/sushy-tools/latest/user/dynamic-emulator.html#virtual-media-resource

MartijnStraatman:
I read something about a storage requirement. Maybe a solution could be this

```
virtctl image-upload pvc live-iso-sno  \
  --size 2Gi --storage-class ocs-storagecluster-ceph-rbd \
  --access-mode ReadWriteOnce \
  --image-path rhcos-4.14.0-x86_64-live.x86_64.iso
```

Storage must be available then through pvc's.

anishbista60:
@starbops I have couple of question. 
- IMO, we need to extend the VM object api to add the `iso` section in the CRD or do we need to implement this in our kubevirtbmc schema ? 
- What sort to field we want to mention in the CR to provide the granular opportunity to custom . Therefore , you can decide the  what backend provider to user and which storage ? 

My rough schema looks like

```yaml
type ISOAttachmentSpec struct {
    // Source of the ISO image - users can choose one of the following
    // Only one source should be specified
    PersistentVolumeClaim *PVCSource    `json:"persistentVolumeClaim,omitempty"`
    URL                   *URLSource     `json:"url,omitempty"`
    ConfigMap             *ConfigMapSource `json:"configMap,omitempty"`
    
    // Storage class to use when creating new PVCs (if needed)
    StorageClassName      *string        `json:"storageClassName,omitempty"`
    
[...]

PR #119, opened 2025-12-04
feat(redfish): support virtual media mount

starbops:
Users can get the virtual media instance, insert media, eject media via the Redfish protocol.

The virtbmc agent creates a DataVolume and then updates the VirtualMachine to perform volume hotplug. Upon media ejection, the VirtualMachine is updated, and the DataVolume is deleted for simplicity.

**Related Issue**

#44 

> [!NOTE]
> ~~Unit tests for the virtualmachine resource manager were removed for the sake of fast prototyping. I will add them back and fix the potential issues later.~~
> Added back.

### DIFF
--- a/pkg/util/helper.go
+++ b/pkg/util/helper.go
@@ -3,6 +3,8 @@ package util
 import (
 	"fmt"
 	"net/http"
+	neturl "net/url"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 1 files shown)


internal/controller/virtualmachinebmc/controller.go
```diff
@@ -256,8 +254,8 @@ func (r *VirtualMachineBMCReconciler) validateVirtualMachineExists(ctx context.C
 //+kubebuilder:rbac:groups=bmc.kubevirt.io,resources=virtualmachinebmcs/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=bmc.kubevirt.io,resources=virtualmachinebmcs/finalizers,verbs=update
 //+kubebuilder:rbac:groups="",resources=pods;services;serviceaccounts,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles,verbs=bind
```
starbops: This is required for the controller manager to create RoleBindings for agents. The referenced ClusterRole contains permissions with CDI DataVolume, which the controller manager's ClusterRole does not possess.

PR #122, opened 2025-12-06
Feat: Refractor the helmchart for v1beta1

anishbista60:
Tracking issue: https://github.com/starbops/kubevirtbmc/issues/113
- After https://github.com/starbops/kubevirtbmc/pull/110, we need change the helm-chart to aligned with current api.


anishbista60:
> @anishbista60 Just out of curiosity, did you ever try the helm plugin for kubebuilder? https://book.kubebuilder.io/plugins/available/helm-v2-alpha

No but looking at docs. If it gonna create helmchart automatically , then it will be very fun 😄

starbops:
I want the project to stick to kubebuilder as closely as possible so we don't have to manually convert the Kustomize manifests to the Helm chart every time. But the feasibility still requires some research.

For this PR, let's leave it as is. I'll do the review.

anishbista60:
I would  say we can review and merge this PR  as of now so  that user won't be affected. 

But future reference: We need to automate this part 

Tracking Issue : http://github.com/starbops/kubevirtbmc/issues/126

PR #124, opened 2025-12-10
Feat: Expose BMC Service to external

anishbista60:
Tracking issue: https://github.com/starbops/kubevirtbmc/issues/3
 
 - Introduced new service field under `spec`
 - Introduced the label and annotation field for that service. 
 - Created the controller loop to handle that new field. 
 - Delete the svc when ever there will be any changes in the svc type.
 - Added the E2E test for this feature leveraging kind loadbalancer.


anishbista60:
@starbops I would say, if you get time just test this part  to check the controller behaviour. I tested it multiple time, it should work.

internal/controller/virtualmachinebmc/controller.go
```diff
@@ -359,6 +397,59 @@ func (r *VirtualMachineBMCReconciler) validateSecretExists(ctx context.Context,
...
+		}
+		log.V(1).Info("created Service for VirtualMachineBMC", "type", desiredType)
+
+	default:
+		log.Error(err, "unable to fetch Service")
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
```
anishbista60: To reduce mess inside the reconcile function, I kinda created this helper one.

I am still in dilema whether to move this part inside actual reconcile function of bmc or  move the service related things under service controller.

anishbista60:
Test case is yet to add but I want to have first look at core functionality.

api/bmc/v1beta1/virtualmachinebmc_types.go
```diff
@@ -30,16 +30,34 @@ const (
...
+
+	// Reference to the VM to manage
+	// +Required
+	VirtualMachineRef *corev1.LocalObjectReference `json:"virtualMachineRef"`
+}
+
+// Service configuration for the BMC service.
+type BMCServiceSpec struct {
+	// Type of service (LoadBalancer, NodePort, ClusterIP, etc)
+	Type corev1.ServiceType `json:"type,omitempty"`
```
starbops: We can leverage the [validation markers](https://book.kubebuilder.io/reference/markers/crd-validation) to protect the field, such as giving it a default value (`ClusterIP`), and restrict the available values.

starbops (commented):
Thanks for the PR!

api/bmc/v1beta1/virtualmachinebmc_types.go
```diff
@@ -52,7 +73,7 @@ type VirtualMachineBMCStatus struct {
 // +kubebuilder:printcolumn:name="VIRTUALMACHINE",type="string",JSONPath=`.spec.virtualMachineRef.name`
 // +kubebuilder:printcolumn:name="SECRET",type="string",JSONPath=`.spec.authSecretRef.name`
 // +kubebuilder:printcolumn:name="CLUSTERIP",type="string",JSONPath=`.status.clusterIP`
```
anishbista60: @starbops I would say, ClusterIp need to replace this with serviceType. What if the service type would be loadbalancer ? Based on this approach , we should also populate the loadbalancer IP. 

If somebody want to see serviceIP, Just see the service ...

starbops: It's fine. We can show the service type, loadbalancer ip, and cluster ip at the same time. If the service type is `ClusterIP`, then the value for `LOADBALANCERIP` will be empty.

internal/controller/service/controller.go
```diff
@@ -75,32 +120,32 @@ func (s *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 
-	// Update VirtualMachineBMC status
-	if svc.Spec.ClusterIP == "" {
-		return ctrl.Result{RequeueAfter: time.Second * 10}, fmt.Errorf("clusterIP is not ready yet")
+	status, svcType := s.checkServiceReadiness(ctx, &svc, &virtualMachineBMC)
+
+	condition := metav1.Condition{
+		Type:    bmcv1.ConditionReady,
+		Status:  metav1.ConditionFalse,
+		Reason:  bmcv1.ConditionNotReady,
```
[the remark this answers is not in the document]
starbops: Could you elaborate?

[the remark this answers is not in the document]
starbops: IIRC a `reason` is mandatory for creating a condition.

api/bmc/v1beta1/virtualmachinebmc_types.go
anishbista60: creating the extra cloumn for servicetype and ip looks queit messy. Current one is fine. Thanks

anishbista60 (commented):
Edited: By mistake

internal/controller/service/controller.go
```diff
@@ -41,6 +41,58 @@ type ServiceReconciler struct {
...
+			status.Message = "ClusterIP assigned to the Service"
+		} else {
+			status.Message = "ClusterIP not yet assigned to the Service"
+		}
+
+	default:
+		log.Error(fmt.Errorf("unsupported service type %s", serviceType), "unsupported Service type for VirtualMachineBMC", "serviceType", serviceType)
+		status.Ready = false
+		status.Message = fmt.Sprintf("unsupported service type: %s", serviceType)
+	}
```
anishbista60: @starbops I think this part is not necessary cause it will be already handled by `	// +kubebuilder:validation:Enum=ClusterIP;NodePort;LoadBalancer
`

```
The VirtualMachineBMC "virtualmachinebmc-sample" is invalid: spec.service.type: Unsupported value: "anish": supported values: "ClusterIP", "NodePort", "LoadBalancer"
```

starbops: No harm to keep it 🙂

anishbista60:
@starbops we should maintain this project for IPMI user as well.

starbops:
@anishbista60 Could you help resolve the conflict? Thank you.

anishbista60:
Since, loadbalancer ip won't be there during e2e  by default. we could leverage this 

https://kind.sigs.k8s.io/docs/user/loadbalancer/

anishbista60:
One thing to note: User who will be using this feature should take care of certificate by their own. We have to mention this in the docs.  Thanks

internal/controller/virtualmachinebmc/controller.go
```diff
@@ -279,16 +280,41 @@ func (r *VirtualMachineBMCReconciler) constructPodFromVirtualMachineBMC(virtualM
...
 
+	labels := map[string]string{
+		VirtualMachineBMCNameLabel: virtualMachineBMC.Name,
+		VMNameLabel:                virtualMachineBMC.Spec.VirtualMachineRef.Name,
+	}
+
+	annotations := map[string]string{}
+
+	if virtualMachineBMC.Spec.Service != nil {
+		if virtualMachineBMC.Spec.Service.Labels != nil {
```
halfcrazy, line 291: Should we preserve system labels `VirtualMachineBMCNameLabel, VMNameLabel` and make them untouchable?

internal/controller/service/controller.go
```diff
@@ -75,32 +127,32 @@ func (s *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		Status:  metav1.ConditionFalse,
+		Reason:  bmcv1.ConditionNotReady,
+		Message: status.Message,
 	}
+
+	if !status.Ready {
+		log.V(1).Info(
+			"Service not ready yet", "serviceType", svcType, "service", svc.Name,
+		)
+		return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
```
halfcrazy, line 162: Service controller returns early when not ready, discarding the prepared 'ServiceNotReady' condition and status changes

internal/controller/service/controller.go
```diff
@@ -75,32 +127,32 @@ func (s *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	} else {
+		condition.Status = metav1.ConditionTrue
+		condition.Reason = bmcv1.ConditionReady
+	}
+
 	if changed := meta.SetStatusCondition(
```
halfcrazy: The reported BMC service IP address stops refreshing once the ready message remains unchanged.
Eg recreate the service manually?

internal/controller/virtualmachinebmc/controller.go
anishbista60: Nice catch. yes, we should preserve that.

anishbista60: But wait re-thinking again. 

When we add a new labels from the BMC CR like 

```yaml
spec
  service: 
    labels: 
      name: anish 
```

The original system labels are preserved. Addition new labels doesn't override it. 

```yaml 
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: "2026-08-03T07:49:12Z"
  labels:
    kubevirt.io/virtualmachinebmc-name: test-bmc
    kubevirt.io/vm-name: testvm
    name: anish
  name: testvm-virtbmc
  namespace: default
  ownerReferences:
  - apiVersion: bmc.kubevirt.io/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: VirtualMachineBMC
    name: test-bmc
    uid: 496056d2-34a3-4bb7-9787-adb54328d0c8
  resourceVersion: "47344"
  uid: 399cae50-16f9-45f1-894c-3bf9eaa01a5b
spec:
  clusterIP: 10.96.178.176
  clusterIPs:
  - 10.96.178.176
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - name: redfish
    port: 80
    protocol: TCP
    targetPort: redfish
  selector:
    kubevirt.io/virtualmachinebmc-name: test-bmc
  sessionAffinity: None
  type: ClusterIP
status:
  loadBalancer: {}
```

internal/controller/service/controller.go
anishbista60: Nice. Yes, status shows the old stale service IP.

internal/controller/service/controller.go
anishbista60: Yes, correct.  Nice catch

internal/controller/virtualmachinebmc/controller.go
halfcrazy: user can use the same key
```
spec
  service: 
    labels: 
      VirtualMachineBMCNameLabel: lorem 
      VMNameLabel : lorem
```

anishbista60: I think this should be fine. As a user, I won't trace in the code and try to override it intentionally the default labels. Instead, I will add the new label that I want to add to that service. 

WDYT ?

anishbista60: those two key are constant: 

https://github.com/kubevirtbmc/kubevirtbmc/blob/main/internal/controller/virtualmachinebmc/constant.go#L16
https://github.com/kubevirtbmc/kubevirtbmc/blob/main/internal/controller/virtualmachinebmc/constant.go#L17

halfcrazy: Since this is quite a tricky scenario, lgtm

halfcrazy (dismissed):

anishbista60:
@halfcrazy thanks for helping with review. 

This is for v0.10.0, let's hold it till then. thank you !

halfcrazy (approved):

starbops:
I will test the PR recently.

BTW, this shouldn’t be put on hold. Once we released v0.9.0, the `main` branch effectively became the development branch for the upcoming v0.10.0 release.

Any fixes we want to include in the `v0.9` branch, especially fixes for bugs found in v0.9.0, should first be made on `main` and then backported to `v0.9` for the v0.9.1 patch release.

anishbista60:
Due to the large merge conflict, this pr is closed by https://github.com/kubevirtbmc/kubevirtbmc/pull/279

Issue #132, opened 2025-12-17
[ENHANCEMENT] Remove hardcoded agent image default and adopt full control via controller flag


anishbista60:
**Is your enhancement request of an existing feature related to a problem? Please describe.**

The agent image has a [hardcoded default in the codebase](https://github.com/starbops/kubevirtbmc/blob/main/internal/controller/virtualmachinebmc/constant.go#L7), even though it [can be overridden via flags](https://github.com/starbops/kubevirtbmc/blob/main/cmd/controller/main.go#L79). This creates unclear precedence and allows the operator to run without explicit image configuration from the deployment.

**Describe the solution you'd like**

Remove the hardcoded default and make `--agent-image-name` and `--agent-image-tag` mandatory flags [at operator deployment](https://github.com/starbops/kubevirtbmc/blob/main/config/manager/manager.yaml#L65). The operator should fail fast at startup if either flag is missing, ensuring the agent image is fully controlled via deployment configuration.

**Additional context**

Current deployments already support to pass the agent image  and tag using flags:

```yaml
args:
  - --agent-image-name= <>
  - --agent-image-tag=<>
```

starbops:
@anishbista60 Thanks for raising this. But I don't think this behavior should be altered. The default values for the agent image name and tag should be hardcoded. The agent stays close to the controller. In most occurrences, their tags (versions) are the same. This provides out-of-the-box ease-of-use experience. If the user wants to use a custom-built image for their agents, they can provide those flags through the controller manager deployment.

What are the benefits of the proposed change?

anishbista60:
> What are the benefits of the proposed change?

The main reason behind this issue was in past development working on so many opened issue, I need to pass the flag every time while working in each PR that was kind of hectic for me.  

I would imagine what if the flag is already there and I just need to change the  `repo name` and `tag`

That's was key behind proposing this issue. May be It might be hectic for me but might not to other's. 

It's more like about Developer Experience :)

anishbista60:
> [@anishbista60](https://github.com/anishbista60) Thanks for raising this. But I don't think this behavior should be altered. The default values for the agent image name and tag should be hardcoded. The agent stays close to the controller. In most occurrences, their tags (versions) are the same. This provides out-of-the-box ease-of-use experience. If the user wants to use a custom-built image for their agents, they can provide those flags through the controller manager deployment.


yeah, I agree with that. We can drop this issue.

anishbista60:
[...]

Issue #40, opened 2024-12-09
[FEATURE] Roll out agents with new images after upgrading the manager


starbops:
**Is your feature request related to a problem? Please describe.**

Existing agents will remain if the user upgrades `virt-controller` with an image other than the previous one. For instance, `virt-controller` was previously deployed with v0.3.1 but then upgraded to v0.3.2; existing agent pods still run in v0.3.1.

**Describe the solution you'd like**

This should require no human intervention. After successfully upgrading the manager, it should automatically roll out new agents. We already expose an option for users to specify what image to use for agents by the manager command, so let's leverage that.

**Describe alternatives you've considered**

Delete the agent pods forcefully after the manager is upgraded and let the reconciliation do its job.

**Additional context**

N/A

anishbista60:
This PR https://github.com/kubevirtbmc/kubevirtbmc/pull/133 will automatically solve this issue.

anishbista60:
Current behaviour automatically solves this issue.

PR #133, opened 2025-12-18
Fix: removed the constant reference to agent image

anishbista60:
Related issue: #132

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -76,8 +76,9 @@ func main() {
 	flag.BoolVar(&secureMetrics, "secure-metrics", true, "Enable secure (HTTPS) metrics endpoint.")
 	flag.BoolVar(&enableHTTP2, "enable-http2", false, "Enable HTTP/2 support (default: disabled for security).")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false, "Enable leader election for controller manager.")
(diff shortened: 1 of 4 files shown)


anishbista60:
@starbops  Taking your points on positional arguments into consideration and thinking about ease of use and operation, I decided to go with **required named flags** for the following reasons:

1.  All major operators use named flags, making this approach familiar to operations teams

2. `--agent-image-name=value` is immediately clear in deployment YAML, while positional values require documentation lookup

3. Named flags prevent accidentally swapping image name and tag values

4. Adding optional flags won't break existing deployments by shifting argument positions

starbops (commented):
I've left a comment on the issue page for better context.

https://github.com/kubevirtbmc/kubevirtbmc/issues/132#issuecomment-3938929519

anishbista60:
@starbops I noticed a useful case for this PR while working on other issue. If we pass the agent name and agent image via flags, it would automatically address #40.

When a user applies the updated single manifest (`kubectl apply -f <latest release>`), the controller manager gets updated. With the agent image provided through the controller manager flags, it will automatically pick up the new image and reconcile existing resources. As a result, the old agent pod will be replaced with a new one using the updated image.

anishbista60:
This is also not worth it to work on for now.

Issue #139, opened 2026-02-14
[BUG] `ipmitool power cycle` does not restart VM – VM remains in Halted state


anishbista60:
## **Describe the bug**

Executing `ipmitool power cycle` against a VM managed by kubevirtbmc does not properly restart the VM.

The current `PowerCycle()` implementation only deletes the `VirtualMachineInstance` (VMI):

```go
func (m *VirtualMachineResourceManager) PowerCycle() error {
	return m.virtClient.KubevirtV1().VirtualMachineInstances(m.namespace).
		Delete(m.ctx, m.name, metav1.DeleteOptions{})
}
```

After deletion of the VMI:

* The VirtualMachine transitions to `Halted`.
* KubeVirt does not automatically recreate the VMI.
* The VM remains stopped instead of performing a power cycle.

## **Expected behavior**

`ipmitool power cycle` should stop and then automatically start the VM, resulting in the VM returning to the `Running` state.

## **Additional context**

`PowerCycle()` currently deletes only the VMI and does not trigger a restart at the VirtualMachine level, which does not align with expected IPMI power cycle semantics.

anishbista60:
This also a bug for Redfish as well.

PR #146, opened 2026-02-23
Fix: trigger VM restart on power cycle call

anishbista60:
Tracking issue: #139 

Note: Functionality is working, I tested it. To remove the effort in manual testing, I am working on writing e2e test currently.


starbops (approved):
LGTM, thank you for the fix.

PR #158, opened 2026-03-28
Feat: changed pod to deployment

anishbista60:
Tracking issue: #162 

- Changed the pod with deployment
- Refractor the controller test 
- Refractor the E2E. 

> [!NOTE]
> As a part this PR the default value of the bmc deployment replica is set to 1. User are not allow to set the replicas to 2 . We need to introduce the new field in the CRD. I will create the sub task for it in issue 136

### DIFF
--- a/internal/controller/virtualmachinebmc/bmc_controller.go
+++ b/internal/controller/virtualmachinebmc/bmc_controller.go
@@ -19,7 +19,6 @@ package virtualmachinebmc
 import (
 	"context"
 	"fmt"
-	"strconv"
 
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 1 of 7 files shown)


anishbista60:
This changes will come as a part of v0.9.0 release

internal/controller/virtualmachinebmc/bmc_controller.go
halfcrazy, line 382: Deployment now

internal/controller/virtualmachinebmc/deployment.go
```diff
@@ -0,0 +1,265 @@
...
+			},
+		},
+	}
+	return deployment
+}
+
+func (r *VirtualMachineBMCReconciler) ensureVirtBMCDeployment(ctx context.Context, virtualMachineBMC *bmcv1.VirtualMachineBMC) error {
+	log := log.FromContext(ctx)
+
+	desiredDeployment := r.createVirtBMCDeployment(virtualMachineBMC)
```
halfcrazy, line 135: Suggest using server-side apply, so we don't need to compare diffs; just apply the patch for all time
```
desired := r.createVirtBMCDeployment(virtualMachineBMC)
if err := ctrl.SetControllerReference(virtualMachineBMC, desired, r.Scheme); err != nil { ... }
  return r.Patch(ctx, desired, client.Apply,
      client.FieldOwner("kubevirtbmc-controller"), client.ForceOwnership)
```

internal/controller/virtualmachinebmc/bmc_controller.go
```diff
@@ -419,20 +318,10 @@ func (r *VirtualMachineBMCReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return ctrl.Result{}, err
 	}
```
halfcrazy, line 320:
```
  // Delete the pre-Deployment fixed-name agent pod so the new Deployment
  // is the sole agent. Runs before Apply to avoid dual-serving on the Service.
  legacyPod := &corev1.Pod{}
  err := r.Get(ctx, types.NamespacedName{Name: deployName, Namespace: ns}, legacyPod)
  if err == nil && metav1.IsControlledBy(legacyPod, virtualMachineBMC) {
      if err := r.Delete(ctx, legacyPod); err != nil && !apierrors.IsNotFound(err) {
          return ctrl.Result{}, err
      }
  } else if err != nil && !apierrors.IsNotFound(err) {
      return ctrl.Result{}, err
  }
```

halfcrazy (changes requested):

anishbista60:
@halfcrazy there are alots of merge conflict. Let me create the another PR with same changes including your suggested changes as well. Thanks

anishbista60:
Closed by https://github.com/kubevirtbmc/kubevirtbmc/pull/276

Issue #131, opened 2025-12-15
[BUG] VirtualMedia.Attach logic creates a PVC with too little available space


sfackler:
**Describe the bug**

When attaching virtual media, the controller looks up the total size of the image and creates a DataVolume requesting a volume with exactly that much storage. However, the storage available to the CDI importer appears to be ~16MiB less than what's requested, so the import fails.

**To Reproduce**

1. Set up a VM and BMC
2. Hit the VirtualMedia.InsertMedia endpoint with an ISO.
3. Observe that the requested storage on the DataVolume that is created is exactly the same size as the ISO.
4. Observe that the CDI import job fails, complaining that the reported available storage is smaller than what was requested.

**Expected behavior**

The DataVolume should successfully populate.

**Screenshots**

```
apiVersion: cdi.kubevirt.io/v1beta1
kind: DataVolume
metadata:
  annotations:
    cdi.kubevirt.io/storage.usePopulator: "true"
  creationTimestamp: "2025-12-15T21:26:05Z"
  generation: 1
  name: testvm
  namespace: vms
  resourceVersion: "1129162578"
  uid: ed82a932-c73e-43b6-b8c5-2d3c6b71a9f4
spec:
  source:
    http:
      url: XXXXXXXXXXXXXXX
  storage:
    resources:
      requests:
        storage: 294220Ki
status:
  claimName: testvm
  conditions:
  - lastHeartbeatTime: "2025-12-15T21:26:38Z"
    lastTransitionTime: "2025-12-15T21:26:05Z"
    message: 'target PVC testvm Pending and [prime-00d276d0-2ed3-4c2e-b1ea-9207818a3ba5]
      : Unable to process data: Unable to transfer source data to scratch space: virtual
      image size 301281280 is larger than the reported available storage 285039616.
      A larger PVC is required'
    reason: Pending
    status: "False"
    type: Bound
  - lastHeartbeatTime: "2025-12-15T21:29:56Z"
    lastTransitionTime: "2025-12-15T21:26:05Z"
    status: "False"
    type: Ready
  - lastHeartbeatTime: "2025-12-15T21:29:56Z"
    lastTransitionTime: "2025-12-15T21:29:56Z"
    message: DataVolume too small to contain image
    reason: Error
    status: "False"
    type: Running
  phase: ImportInProgress
  progress: N/A
  restartCount: 5
```

[...]

PR #170, opened 2026-04-27
fix: add 10% storage overhead to DataVolume to account for CDI filesystem overhead

anishbista60:
Tracking issue: #131 

When attaching virtual media, the DataVolume was created with storage exactly equal to the image size. CDI's scratch PVC has filesystem overhead (nearly 5-10%) which reduces usable space below the image size, causing the import to fail.

Adding 10% padding to the storage request, which covers CDI's default 6% overhead with a safe buffer.

Testing: 

Before: 

```yaml

  capacity:
    storage: "3502306755"
  phase: Bound
```

After adding `0.10` overhead

```yaml

status:
  capacity:
    storage: "3852426281"
  phase: Bound
```


anishbista60:
We do not need to fix from our end.Closing this 

Adding the docs for it: https://github.com/kubevirtbmc/docs/pull/5

Issue #173, opened 2026-04-27
[BUG] EjectMedia returns 400 on empty request body sent by sushy


anishbista60:
**Description**

Sushy calls `EjectMedia` with an empty HTTP body during `prepare_ramdisk` and `clean_up_ramdisk`. The generated OpenAPI handler requires a JSON body and returns `400 Bad Request` with body `"EOF"`. Sushy then crashes trying to parse that string as JSON.

**Error from Ironic**
```
unable to start inspection: 'str' object has no attribute 'get'
```

**Root cause**

In `pkg/generated/redfish/server/api_default.go`, the `EjectMedia` handler uses `json.NewDecoder(r.Body).Decode(&bodyParam)` which returns `io.EOF` for an empty body, treated as a parsing error.

**Fix**

Modify 4 EjectMedia handler functions to tolerate empty bodies:
```go
// Before:
if err := d.Decode(&bodyParam); err != nil {
// After:
if err := d.Decode(&bodyParam); err != nil && err != io.EOF {
```

starbops:
I suppose this should be due to an older version behavior of sushy?

According to the Redfish specification[^1], the client should send an empty JSON object:

<img width="951" height="806" alt="Image" src="https://github.com/user-attachments/assets/1033d6b7-cd44-475c-b850-0555ec99b65d" />

@anishbista60 What's the version of sushy you used?

[^1]: https://www.dmtf.org/sites/default/files/standards/documents/DSP0266_1.20.0.html#post-action

anishbista60:
Older version of sushy ? 
Client could be any version right ? The request needs to be handled correctly by the redfish server.

starbops:
I suppose this should be bug on the sushy side. KubeVirtBMC follows the Redfish spec. Specifically, if a POST request does not have any required parameter, the client should send an empty JSON object instead of an empty string in the request body.

anishbista60:
> I suppose this should be bug on the sushy side. KubeVirtBMC follows the Redfish spec. Specifically, if a POST request does not have any required parameter, the client should send an empty JSON object instead of an empty string in the request body.

Hmm, Since the issue was filed based out of metal 3 folks. I asked him about the sushy version then only I can confirm.

anishbista60:
But one more thing if you see the screen shot in the linked PR, i faced same issue from the redfish pod as well

lentzi90:
[...]

PR #180, opened 2026-05-02
  Fix: tolerate empty body in EjectMedia handler for VirtualMedia

anishbista60:
Tracking issue: #173 

## Before Fix: 

Screenshot From 2026-05-02 11-26-42

## After fix: 

Screenshot From 2026-05-02 11-20-47

### DIFF
--- a/pkg/generated/redfish/server/api_default.go
+++ b/pkg/generated/redfish/server/api_default.go
@@ -13,6 +13,7 @@ package server
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


starbops (commented):
Thanks for the PR. I'm fine tolerating the EjectMedia requests with empty body, if that facilitate the sushy integration.

Can we add a noting comment explain why we did this for this specific API endpoint? And we need a test case for this, specifically at the agent EE testing.

anishbista60:
> Thanks for the PR. I'm fine tolerating the EjectMedia requests with empty body, if that facilitate the sushy integration.

Can you check the message in the same gh issue ? 

This PR doesn't seems worthy at a moment. We can again come back to solve this issue as a patch once we get response from metal 3 folks.

starbops (commented):
I'm fine with this, but could you add a comment explaining that this is done for the convenience of various client implementations, so future readers understand the rationale?

Thank you.

starbops (approved):
Awesome. LGTM, thank you for the fix!

starbops:
@mergify backport v0.8

Issue #175, opened 2026-04-27
[BUG] PowerOn races CDI DataVolume import causing boot failure


anishbista60:
**Description**

There is a timing mismatch between Ironic's synchronous virtual media expectations and KubeVirtBMC's asynchronous CDI-based implementation. `InsertMedia` creates a CDI DataVolume and returns immediately. Ironic then calls `PowerOn`, and the VM boots before CDI finishes downloading the IPA deploy ISO. Ironic eventually times out, deletes the ISO from httpd, and CDI gets a 404.

**Root cause**

CDI with `WaitForFirstConsumer` storage classes won't start importing until a consumer pod exists. The VM start creates the virt-launcher pod, which triggers PVC binding, which triggers CDI import — but this all happens after `InsertMedia` has already returned.

**Fix**

Add a polling loop in `PowerOn()` after setting `runStrategy` that waits for all hotpluggable DataVolume-backed volumes to reach `Succeeded` phase before returning. Timeout: 10 minutes, poll interval: 5 seconds.

```go
// In PowerOn(), after updating the VM's runStrategy:
if err := m.waitForDataVolumes(vm); err != nil {
    return err
}
```

**Boot sequence after fix**
```
1. InsertMedia   → creates DV, attaches to VM, returns immediately
2. SetBootDevice → sets CDROM boot order
3. PowerOn       → sets runStrategy → PVC binds → CDI imports → waits → returns
4. Ironic        → waits for IPA callback (media is ready)
```

anishbista60:
The issue is with the storage class which blocks the DataVolume in `WaitForFirstConsumer` state when you run the media mount for stopped vm. 

The simplest fix would be to bypass the ISO download part using data volume annotation and keep the rest as-is which are set the CD-ROM as the boot device and power on or restart the VM 

In the attached PR, I have not used any polling to make `InsertMedia` wait until the ISO image download is completed. Should we go with polling inside `InsertMedia`, or leave it to the user to write their own wait logic for the DataVolume?

Now, In this case if media mount return immediately, don't call the power on by any tool like sushy e.t.c until iso is downloaded, to fill that time gap user needs to write few line of extra code to wait for data volume to reach to completed state. 

WDYT @starbops

starbops:
@anishbista60 Sorry I missed this one.

[...]

PR #186, opened 2026-05-24
fix: add  annotation to DataVolume for WaitForFirstConsumer support

anishbista60:
Tracking issue: #175 

### Before
 - If the StorageClass had volumeBindingMode: WaitForFirstConsumer, the PVC remained Pending on a stopped VM (no consumer to trigger binding), so the ISO download never started.
 
### After: 
- The annotation cdi.kubevirt.io/storage.bind.immediate.requested is set on the DataVolume, telling CDI to bind the PVC immediately regardless of WaitForFirstConsumer. The importer pod runs and downloads the ISO even on a stopped VM.

### DIFF
--- a/pkg/util/helper.go
+++ b/pkg/util/helper.go
@@ -13,6 +13,10 @@ import (
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 )
 
+// AnnStorageBindImmediateRequested is the DataVolume annotation that requests
+// immediate binding of the underlying PVC, bypassing WaitForFirstConsumer.
+const AnnStorageBindImmediateRequested = "cdi.kubevirt.io/storage.bind.immediate.requested"
+
 func Ptr[T any](value T) *T {
 	return &value
 }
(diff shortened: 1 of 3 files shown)


starbops (commented):
Could you help rebase the PR to the latest main branch?

pkg/util/helper.go
```diff
@@ -54,6 +54,9 @@ func ConstructDataVolume(namespace, name, url string, size int64) *cdiv1.DataVol
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: namespace,
 			Name:      name,
+			Annotations: map[string]string{
+				"cdi.kubevirt.io/storage.bind.immediate.requested": "",
```
starbops: Please make the annotation key as a constant.

anishbista60:
This require e2e. I will add it while working with #205

anishbista60:
But I have tested it locally, therefore feel free to move ahead.

starbops (approved):
LGTM, thank you!

starbops:
@mergify rebase v0.8

starbops:
@mergify backport v0.8

PR #187, opened 2026-06-08
feat: Use KubeVirt subresource APIs for power operations

### LINKED ISSUE #190: [ENHANCEMENT] Use KubeVirt subresource APIs for power operations

We have our own fork of kubevirt have some extra fields for vm cr, when using kubevirtbmc to do power operations, our own fields were stripped. So I'm trying not to update the whole vm cr, better to use a subresource or patch the specific field.

Not to update the whole vm cr, better to use a subresource or patch the specific field.

**Additional context**
I'm using the fork version of kubevirt.

halfcrazy:
Replace GET→modify→Update(PUT) pattern in PowerOn/PowerOff with KubeVirt's Start/Stop subresource APIs. This avoids race conditions and respects KubeVirt's state machine for VM lifecycle transitions.

- PowerOn: Use virtClient.KubevirtV1().VirtualMachines().Start()
- PowerOff: Use virtClient.KubevirtV1().VirtualMachines().Stop()
- PowerCycle: Use Start() subresource for halted VMs (instead of mutating spec via PowerOn), keep Restart() for running VMs
- Handle 'already running'/'already stopped' errors as success (idempotent operations that achieved the desired state)
- Add error logging in IPMI handler before returning completion codes
- Add RBAC permissions for subresources.kubevirt.io API group
- Update tests to verify operation success

fix #190

### DIFF
--- a/pkg/ipmi/handler.go
+++ b/pkg/ipmi/handler.go
@@ -38,6 +38,7 @@ func (h *handler) chassisControlHandler(m *goipmi.Message) goipmi.Response {
 	}
 
 	if err != nil {
+		logrus.WithError(err).Error("chassis control failed")
 		return &goipmi.ChassisControlResponse{
 			CompletionCode: goipmi.ErrInvalidState,
 		}
@@ -53,6 +54,7 @@ func (h *handler) chassisStatusHandler(*goipmi.Message) goipmi.Response {
 
 	isUp, err := h.rm.GetPowerStatus()
 	if err != nil {
+		logrus.WithError(err).Error("get power status failed")
 		return &goipmi.ChassisStatusResponse{
 			CompletionCode: goipmi.ErrInvalidState,
 		}
--- a/pkg/resourcemanager/virtualmachine_resourcemanager.go
+++ b/pkg/resourcemanager/virtualmachine_resourcemanager.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/sirupsen/logrus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kubevirtv1 "kubevirt.io/api/core/v1"
 	cdiclient "kubevirt.io/client-go/containerizeddataimporter"
@@ -252,31 +253,44 @@ func (m *VirtualMachineResourceManager) GetPowerStatus() (bool, error) {
 }
 
 func (m *VirtualMachineResourceManager) PowerOn() error {
-	err := m.virtClient.KubevirtV1().VirtualMachines(m.namespace).
-		Start(m.ctx, m.name, &kubevirtv1.StartOptions{})
+	exists, err := m.isVMIExists()
 	if err != nil {
-		// The VM is already running — the operation succeeded.
-		if strings.Contains(err.Error(), "already running") {
-			logrus.WithError(err).Info("VM is already running")
-			return nil
-		}
-		return err
+		return err // propagate transient API errors
 	}
-	return nil
+	if exists {
+		logrus.Info("VM is already running, skipping Start")
+		return nil
+	}
+	return m.virtClient.KubevirtV1().VirtualMachines(m.namespace).
+		Start(m.ctx, m.name, &kubevirtv1.StartOptions{})
 }
 
 func (m *VirtualMachineResourceManager) PowerOff() error {
-	err := m.virtClient.KubevirtV1().VirtualMachines(m.namespace).
-		Stop(m.ctx, m.name, &kubevirtv1.StopOptions{})
+	exists, err := m.isVMIExists()
 	if err != nil {
-		// The VM is already stopped — the operation succeeded.
-		if strings.Contains(err.Error(), "already stopped") {
-			logrus.WithError(err).Info("VM is already stopped")
-			return nil
-		}
-		return err
+		return err // propagate transient API errors
 	}
-	return nil
+	if !exists {
+		logrus.Info("VM is already stopped, skipping Stop")
+		return nil
+	}
+	return m.virtClient.KubevirtV1().VirtualMachines(m.namespace).
+		Stop(m.ctx, m.name, &kubevirtv1.StopOptions{})
+}
+
+// isVMIExists checks whether a VirtualMachineInstance exists for this VM.
+// Returns (true, nil) if the VMI exists, (false, nil) if it doesn't,
+// or (false, error) if the API call failed for any other reason.
+func (m *VirtualMachineResourceManager) isVMIExists() (bool, error) {
+	_, err := m.virtClient.KubevirtV1().VirtualMachineInstances(m.namespace).
+		Get(m.ctx, m.name, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
 }
 
 func (m *VirtualMachineResourceManager) PowerCycle() error {
--- a/pkg/resourcemanager/virtualmachine_resourcemanager.go
+++ b/pkg/resourcemanager/virtualmachine_resourcemanager.go
@@ -253,44 +253,29 @@ func (m *VirtualMachineResourceManager) GetPowerStatus() (bool, error) {
 }
 
 func (m *VirtualMachineResourceManager) PowerOn() error {
-	exists, err := m.isVMIExists()
-	if err != nil {
-		return err // propagate transient API errors
-	}
-	if exists {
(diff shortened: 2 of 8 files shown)


anishbista60 (commented):
Hey, Thanks for PR. Using sub-resource api directly is much cleaner. Could you please create the issue to track this PR  as a enhancement ? 

Couple of question/suggestion.

pkg/resourcemanager/virtualmachine_resourcemanager.go
```diff
@@ -252,46 +252,28 @@ func (m *VirtualMachineResourceManager) GetPowerStatus() (bool, error) {
...
-		}(kubevirtv1.RunStrategyRerunOnFailure)
-		vm.Spec.RunStrategy = runStrategy
-	}
-	if _, err := m.virtClient.KubevirtV1().VirtualMachines(m.namespace).
-		Update(m.ctx, vm, metav1.UpdateOptions{}); err != nil {
+		// The VM is already running — the operation succeeded.
+		if strings.Contains(err.Error(), "already running") {
+			logrus.WithError(err).Info("VM is already running")
+			return nil
+		}
```
anishbista60: I would recommend to check status of VMI to see whether it is running or not rather than relying on string comparison

pkg/resourcemanager/virtualmachine_resourcemanager.go
```diff
@@ -252,46 +252,28 @@ func (m *VirtualMachineResourceManager) GetPowerStatus() (bool, error) {
...
-		}(kubevirtv1.RunStrategyHalted)
-		vm.Spec.RunStrategy = runStrategy
-	}
-	if _, err := m.virtClient.KubevirtV1().VirtualMachines(m.namespace).
-		Update(m.ctx, vm, metav1.UpdateOptions{}); err != nil {
+		// The VM is already stopped — the operation succeeded.
+		if strings.Contains(err.Error(), "already stopped") {
+			logrus.WithError(err).Info("VM is already stopped")
+			return nil
+		}
```
anishbista60: Ditto

anishbista60:
Also, we need to change the role for the helmchart as well. Please address that as well as the part of this PR.

pkg/ipmi/handler.go
```diff
@@ -98,6 +100,7 @@ func (h *handler) setSystemBootOptionsHandler(m *goipmi.Message) goipmi.Response
 
 	err := h.rm.SetBootDevice(device)
 	if err != nil {
+		logrus.WithError(err).Error("set boot device failed")
```
anishbista60: Seems, this is not relevant in this PR.

pkg/resourcemanager/virtualmachine_resourcemanager.go
```diff
@@ -252,46 +252,28 @@ func (m *VirtualMachineResourceManager) GetPowerStatus() (bool, error) {
 }
 
 func (m *VirtualMachineResourceManager) PowerOn() error {
-	vm, err := m.virtClient.KubevirtV1().VirtualMachines(m.namespace).
-		Get(m.ctx, m.name, metav1.GetOptions{})
+	err := m.virtClient.KubevirtV1().VirtualMachines(m.namespace).
+		Start(m.ctx, m.name, &kubevirtv1.StartOptions{})
```
anishbista60: I would say, for this powerOn() function correct order of code would be first check whether VMI is running or not. If not then call start subresource api. 

Meaning, https://github.com/kubevirtbmc/kubevirtbmc/pull/187#discussion_r3371177756 part will come at first then this part will come later

anishbista60: Also for poweroff() function as well.

halfcrazy:
> Also, we need to change the role for the helmchart as well. Please address that as well as the part of this PR.

done

pkg/resourcemanager/virtualmachine_resourcemanager.go
halfcrazy: Kubevirt uses the same methods to check whether a VM is running. I think it's not the problem.(virt-api will handle the internal status check)
https://github.com/kubevirt/kubevirt/blob/540f37cfa3caed601ac6710cb002b11a249d0639/tests/compute/vm_lifecycle.go#L243-L257

anishbista60: The KubeVirt test you referenced is asserting that the error message is stable from KubeVirt's own test suite, they own that string. kubevirtbmc is an external consumer, so relying on that error message for control flow is fragile

What if they change that wording in next release ? At that time it will break

pkg/resourcemanager/virtualmachine_resourcemanager.go
```diff
@@ -252,49 +253,44 @@ func (m *VirtualMachineResourceManager) GetPowerStatus() (bool, error) {
...
 }
 
 func (m *VirtualMachineResourceManager) PowerOff() error {
-	vm, err := m.virtClient.KubevirtV1().VirtualMachines(m.namespace).
-		Get(m.ctx, m.name, metav1.GetOptions{})
+	exists, err := m.isVMIExists()
 	if err != nil {
-		return err
+		return err // propagate transient API errors
```
anishbista60:
```suggestion
		return err
```

pkg/resourcemanager/virtualmachine_resourcemanager.go
```diff
@@ -252,49 +253,44 @@ func (m *VirtualMachineResourceManager) GetPowerStatus() (bool, error) {
 }
 
 func (m *VirtualMachineResourceManager) PowerOn() error {
-	vm, err := m.virtClient.KubevirtV1().VirtualMachines(m.namespace).
-		Get(m.ctx, m.name, metav1.GetOptions{})
+	exists, err := m.isVMIExists()
 	if err != nil {
-		return err
+		return err // propagate transient API errors
```
anishbista60:
```suggestion
		return err 
```

pkg/resourcemanager/virtualmachine_resourcemanager.go
```diff
@@ -252,49 +253,44 @@ func (m *VirtualMachineResourceManager) GetPowerStatus() (bool, error) {
...
+		Get(m.ctx, m.name, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
 	}
-	return nil
+	return true, nil
 }
```
anishbista60: I would say to inline `isVMIExists`  function inside poweron(). This way no need to create another function.

something like this


```go
func (m *VirtualMachineResourceManager) PowerOn() error {
    _, err := m.virtClient.KubevirtV1().VirtualMachineInstances(m.namespace).
        Get(m.ctx, m.name, metav1.GetOptions{})
    if err != nil && !apierrors.IsNotFound(err) {
        return fmt.Errorf("failed to get VMI %s/%s: %w", m.namespace, m.name, err)
    }
    if err == nil {
        return nil
    }
    return m.virtClient.KubevirtV1().VirtualMachines(m.namespace).
        Start(m.ctx, m.name, &kubevirtv1.StartOptions{})
}

```

anishbista60 (commented):
Rest LGTM. /cc @starbops

anishbista60 (dismissed):
LGTM. Thank you for the PR. Let's see the CI now. For Final merge: @starbops please take a look.

anishbista60:
Please check the e2e failure.

halfcrazy:
> Please check the e2e failure.

Cannot reproduce in the local cluster. Maybe a flake case. Could you re-run the failed action?

test/virtbmc-agent/agent_helpers.go
```diff
@@ -306,7 +306,7 @@ func runCurlRedfish(ctx context.Context, cfg *rest.Config, namespace string, r R
 	if r.Path != "" {
 		url = strings.TrimSuffix(r.BaseURL, "/") + r.Path
 	}
-	args := []string{"-i", "-L", "-X", r.Method}
+	args := []string{"--connect-timeout", "5", "--max-time", "15", "-i", "-L", "-X", r.Method}
```
anishbista60, line 309: nit: One last question, since, we you already added readiness probe to check the redfish server to be ready. 

Is this really needed now ?

halfcrazy: The readiness probe handles "is the server up before we start," while the curl timeout handles "don't let a single stuck request hang the test."

anishbista60 (approved):
LGTM from my side. Thanks for the enhancement. 

Final approve and merge: Please @starbops take a look. Thanks !

starbops (approved):
@halfcrazy Thank you for the enhancement. LGTM!

PR #188, opened 2026-06-08
feat: Use JSON Patch for boot device operations with full device sequencing

### LINKED ISSUE #184: [BUG] `SetBootDevice` blindly uses index 0 for HDD and PXE boot device selection
## Problem

`SetBootDevice` uses hardcoded index 0 when setting boot order for 
`BootDeviceHdd` and `BootDevicePxe`, without checking what type of 
device is actually at that position.

### Affected Code
- `BootDeviceHdd` → `Disks[0]` may be a CD-ROM depending on VM spec ordering
- `BootDevicePxe` → `Interfaces[0]` may not be the PXE-bootable interface

## Impact
- Requesting HDD boot may silently set boot order on a CD-ROM disk
- Requesting PXE boot may silently set boot order on the wrong interface
- No error is returned - wrong behaviour is silent

## Expected Behaviour
Search through the list by device type, not by position, consistent with how `BootDeviceCd` already works via `GetCdromDisk`.

halfcrazy:
fix #184 

Replace the GET→modify→Update(PUT) pattern in SetBootDevice with targeted JSON Patch operations. Each device's bootOrder is set via individual 'add' patch op, preserving all other VM properties.

Also implements proper IPMI boot order semantics by sequencing ALL devices by category with the chosen category first:

  PXE:   NICs(1,2,3) → Disks(4,5,6) → CDROMs(7,8,9)
  HDD:   Disks(1,2,3) → NICs(4,5,6) → CDROMs(7,8,9)
  CDROM: CDROMs(1,2,3) → Disks(4,5,6) → NICs(7,8,9)

Devices are classified as: CDROM (disk.CDRom != nil), regular disk (Disk, LUN, Floppy), or NIC (interface). JSON Patch 'add' operation is used (not 'replace') to handle both existing and absent bootOrder fields.

### DIFF
--- a/pkg/resourcemanager/virtualmachine_resourcemanager.go
+++ b/pkg/resourcemanager/virtualmachine_resourcemanager.go
@@ -2,6 +2,7 @@ package resourcemanager
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"slices"
 	"strings"
@@ -9,6 +10,7 @@ import (
 	"github.com/sirupsen/logrus"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	kubevirtv1 "kubevirt.io/api/core/v1"
 	cdiclient "kubevirt.io/client-go/containerizeddataimporter"
 	kvclient "kubevirt.io/client-go/kubevirt"
@@ -291,7 +293,10 @@ func (m *VirtualMachineResourceManager) PowerCycle() error {
 }
 
 func (m *VirtualMachineResourceManager) SetBootDevice(bootDevice BootDevice) error {
-	logrus.Info("SetBootDevice")
+	logrus.Infof("SetBootDevice: %s", bootDevice)
+
+	// Fetch the VM only to discover device indices for patch paths.
+	// The actual mutation is done via JSON Patch to avoid full-UPDATE races.
 	vm, err := m.virtClient.KubevirtV1().VirtualMachines(m.namespace).
 		Get(m.ctx, m.name, metav1.GetOptions{})
 	if err != nil {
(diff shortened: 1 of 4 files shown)


anishbista60:
@halfcrazy could you please rebase this PR with the main ?

halfcrazy:
> @halfcrazy could you please rebase this PR with the main ?

done

starbops:
@halfcrazy Thank you for the PR. We'll check it ASAP. Also, could you please leave a comment in #184 so that we can assign the issue to you?

anishbista60 (commented):
Thanks for the fix and enhancement at the same time @halfcrazy. I don't see much suggestion in code quality. Json patch is way to go 🥇 

Overall looks good to me.

test/virtbmc-agent/agent_test.go
```diff
@@ -238,6 +253,11 @@ var _ = Describe("Agent e2e", Ordered, func() {
 					ContainSubstring("200"),
 					ContainSubstring("204"),
 				))
+				// PXE: interfaces first, then regular disks, then cdroms
+				verifyVMBootOrder(ctx, k8sClient, ns,
+					map[int]uint{0: 2, 1: 3}, // disks: regular=2, cdrom=3
+					map[int]uint{0: 1},       // interface=1
+				)
```
anishbista60, line 260: Thanks for improving the e2e because it was pending for long time. Previously, It was just checking from client side and now this test cover the validation from VM side as well.

anishbista60:
I would still like to test all these behaviour in my local by today then I will approve it. Thanks !

pkg/resourcemanager/virtualmachine_resourcemanager.go
```diff
@@ -302,47 +307,75 @@ func (m *VirtualMachineResourceManager) SetBootDevice(bootDevice BootDevice) err
...
-		if vm.Spec.Template.Spec.Domain.Devices.Interfaces == nil {
-			return fmt.Errorf("no interfaces found")
+		if len(ifaces) == 0 {
+			return fmt.Errorf("no interfaces found for PXE boot")
+		}
+		ordered = []deviceGroup{
+			{"interface", "/spec/template/spec/domain/devices/interfaces", ifaceIndices},
+			{"disk", "/spec/template/spec/domain/devices/disks", regularDiskIndices},
+			{"disk", "/spec/template/spec/domain/devices/disks", cdromIndices},
 		}
```
starbops, line 343: If the boot device is set to PXE, the network interface group should clearly be placed ahead of the other two groups. My question is how we should determine the ordering of the remaining groups. Does it always have to follow a fixed order?

I think both of the following approaches would work. I'm just wondering whether there is any rationale for preferring one over the other.

```
ordered = []deviceGroup{
	{"interface", "/spec/template/spec/domain/devices/interfaces", ifaceIndices},
	{"disk", "/spec/template/spec/domain/devices/disks", regularDiskIndices},
	{"disk", "/spec/template/spec/domain/devices/disks", cdromIndices},
}
```

```
ordered = []deviceGroup{
	{"interface", "/spec/template/spec/domain/devices/interfaces", ifaceIndices},
	{"disk", "/spec/template/spec/domain/devices/disks", cdromIndices},
	{"disk", "/spec/template/spec/domain/devices/disks", regularDiskIndices},
}
```

starbops (commented):
I've tested with some basic operations and it works perfectly. Only one question. Please see below. Thank you for the PR.

pkg/resourcemanager/virtualmachine_resourcemanager.go
halfcrazy: In the PXE boot scenario, the hard drive has a higher boot priority than the optical drive.
The main reason is device positioning. Local hard drives carry the resident OS and are the primary boot device for normal operation. Optical drives are auxiliary devices, mainly used for one-time tasks like system installation, troubleshooting and data reading. For this practical consideration, BIOS/UEFI sets the hard drive to boot before the optical drive after PXE.

starbops: Make sense. Thank you for elaborating.

LGTM!

starbops (approved):

anishbista60 (approved):
Alright, cool. Thanks @halfcrazy for your good work. 

Thank you @starbops  for testing it in your local. 

Way to `v0.8.0` release.

anishbista60:
@halfcrazy BTW, We will make new release soon cause all the issue has been sorted for `v0.8.0`. If it is ok for you, please drop your Linkedin profile over here. 

I would like to give shoutout/thanks to you when I will make the release announcement in Linkedin. 

Thanks !

halfcrazy:
> @halfcrazy BTW, We will make new release soon cause all the issue has been sorted for `v0.8.0`. If it is ok for you, please drop your Linkedin profile over here.
> 
> I would like to give shoutout/thanks to you when I will make the release announcement in Linkedin.
> 
> Thanks !

Thank you. My LinkedIn profile https://www.linkedin.com/in/%E5%BD%A6-%E6%9C%B1-40b20591/

PR #193, opened 2026-06-11
chore: enable KubeVirt useEmulation in e2e install

### LINKED ISSUE #196: [ENHANCEMENT] E2E: using emulation for kubevirt when kvm is not available

Running e2e local requires KVM support; otherwise, the VM will not run.

Setting up kubevirt using emulation will solve this issue. https://github.com/kubevirt/kubevirt/blob/main/docs/software-emulation.md

**Additional context**

Local machine doesn't support KVM.

halfcrazy:
Patches the KubeVirt CR after apply to enable software emulation, allowing VMs to run in environment without hardware virtualization.

fix #196

### DIFF
--- a/test/util/helper.go
+++ b/test/util/helper.go
@@ -128,6 +128,13 @@ func InstallKubeVirt() error {
 		return fmt.Errorf("apply KubeVirt CR: %w", err)
 	}
 
+	// Enable emulation so VMs can run without hardware virtualization (e.g. in Kind).
+	cmd = exec.Command("kubectl", "patch", "kubevirt", "kubevirt", "-n", "kubevirt", "--type=merge",
+		"-p", `{"spec":{"configuration":{"developerConfiguration":{"useEmulation":true}}}}`)
+	if _, err := Run(cmd); err != nil {
+		return fmt.Errorf("patch KubeVirt CR for useEmulation: %w", err)
+	}
+
 	Eventually(func() (string, error) {
 		cmd := exec.Command("kubectl", "get", "kubevirt", "kubevirt", "-n", "kubevirt", "-o", "jsonpath={.status.phase}")
 		out, err := Run(cmd)
--- a/test/util/helper.go
+++ b/test/util/helper.go
@@ -39,6 +39,11 @@ func Run(cmd *exec.Cmd) (string, error) {
 	return string(out), nil
 }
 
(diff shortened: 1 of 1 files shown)


test/util/helper.go
```diff
@@ -128,6 +128,13 @@ func InstallKubeVirt() error {
 		return fmt.Errorf("apply KubeVirt CR: %w", err)
 	}
 
+	// Enable emulation so VMs can run without hardware virtualization (e.g. in Kind).
+	cmd = exec.Command("kubectl", "patch", "kubevirt", "kubevirt", "-n", "kubevirt", "--type=merge",
+		"-p", `{"spec":{"configuration":{"developerConfiguration":{"useEmulation":true}}}}`)
+	if _, err := Run(cmd); err != nil {
+		return fmt.Errorf("patch KubeVirt CR for useEmulation: %w", err)
+	}
+
```
anishbista60: Hey, Thanks for the PR. Since the e2e runs on the GitHub runner, which supports KVM virtualization, this won't be required for the CI part. 

Now, coming to the local e2e, if your laptop is lfor eg:  Mac which does not support KVM, then it requires software emulation to be set to true.

But in that case, the developer can do it themselves in their testing environment rather than adding it to e2e, because this e2e will run as part of the CI and software emulation is very slow.

halfcrazy: Emm, `make local-e2e-test` has no chance to inject this configuration. I have to open another window and wait for the controller e2e start.
How about adding a check for KVM; if not satisfied, use emulation?

anishbista60: yep, that's sounds good. Please create issue to track this.

test/util/helper.go
```diff
@@ -128,6 +133,18 @@ func InstallKubeVirt() error {
 		return fmt.Errorf("apply KubeVirt CR: %w", err)
 	}
 
+	// Enable emulation only when KVM is not available (e.g. in Kind).
```
anishbista60: Nit: 
```suggestion
	// Enable emulation only when KVM is not available.
```
There is nothing to do with any type of kubernetes. Only deal is whether that machine support KVM or not.

anishbista60 (approved):
LGTM. Thanks for enhancement !

starbops (approved):
LGTM, thank you for the enhancement!

Issue #30, opened 2024-10-23
[FEATURE] RMCP+ support


starbops:
**Is your feature request related to a problem? Please describe.**

Currently, for the IPMI-over-LAN functionality, KubeVirtBMC only supports RMCP. But nowadays the mainstream is RMCP+, which is with the IPMIv2.

```
# IPMI 2.0
$ ipmitool -I lanplus -U admin -P password -H  172.19.31.70 power status
Error: Unable to establish IPMI v2 / RMCP+ session

# IPMI 1.5
$ ipmitool -I lan -U admin -P password -H  172.19.31.70 power status
Chassis Power is on
```

For example, [bmclib](https://github.com/bmc-toolbox/bmclib) does not support IPMI v1.5 at all. The flag `-I lanplus` is hardcoded in every `ipmitool` commands executed: https://github.com/bmc-toolbox/bmclib/blob/8a5f5a1a243c8f6174bb139a45c6cba30dabff9c/internal/ipmi/ipmi.go#L77

On the other hand, Tinkerbell Rufio leverages bmclib to communicate with target machine's BMC. As a result, KubeVirtBMC does not play well with Rufio. You'll get something like the following when adding a new Machine object:

```
{"level":"info","controller":"machine","controllerGroup":"bmc.tinkerbell.org","controllerKind":"Machine","Machine":{"name":"machine-1","namespace":"tink-system"},"namespace":"tink-system","name":"machi
ne-1","reconcileID":"9e871bba-03b2-42f2-bdb9-99b36ef83783","host":"172.19.31.70","username":"admin","v":0,"logger":"controllers/Machine","error":"8 errors occurred:\n\t* provider: asrockrack: failed to
 login\n\t* provider: supermicro: 404: failed to login\n\t* provider: gofish: 404: 404 page not found\n\n\t* provider: dell: 404: 404 page not found\n\n\t* provider: openbmc: not an OpenBMC device\n\t*
 provider: ipmitool: Error: Unable to establish IPMI v2 / RMCP+ session: exit status 1\n\t* provider: IntelAMT: unable to perform digest auth with http://172.19.31.70:16992/wsman: Post \"http://172.19.
31.70:16992/wsman\": context deadline exceeded\n\t* no Opener implementations found\n\n","providersAttempted":["gofish","dell","openbmc","ipmitool","asrockrack","IntelAMT","supermicro"],"successfulProv
[...]

PR #194, opened 2026-06-11
fix: ipmi authorization

### LINKED ISSUE #195: [BUG] IPMI doesn't check authentication
**Describe the bug**

IPMI doesn't check authentication: both username and password.

**To Reproduce**
```
/ # ipmitool -I lan -U admin -P password -H vm-kfgk3ksdlmk4mcol1uop-virtbmc.default.svc.cluster.local power status
Chassis Power is off

/ # ipmitool -I lan -U admin -P 123456 -H vm-kfgk3ksdlmk4mcol1uop-virtbmc.default.svc.cluster.local power status
Chassis Power is off

/ # ipmitool -I lan -U lorem -P 123456 -H vm-kfgk3ksdlmk4mcol1uop-virtbmc.default.svc.cluster.local power status
Chassis Power is off
```

```
[root@xxx ~]# kubectl get secret bmc-secret -oyaml
apiVersion: v1
data:
  password: YWRtaW4xMjM=
  username: YWRtaW4=
kind: Secret
metadata:
  creationTimestamp: "2026-06-02T07:02:18Z"
  name: bmc-secret
  namespace: default
  resourceVersion: "185693756"
  uid: 0312544f-c8f1-4196-8543-e86604f360d8
type: Opaque
[root@xxx ~]# echo YWRtaW4= |base64 -d
admin
[root@xxx ~]# echo YWRtaW4xMjM= |base64 -d
admin123
```

**Expected behavior**

An incorrect username or password should failed authorization.

**Screenshots**

- OS: [e.g. SLES 15 SP5]
- Chart version: [e.g. 0.2.3]
- App version: [e.g. v0.5.3]

**Additional context**

halfcrazy:
The current IPMI implementation does not support password verification, and passwords configured in the secret take no effect. This MR adds the verification logic. The protocol interaction part was developed largely with AI coding tools.

fix https://github.com/kubevirtbmc/kubevirtbmc/issues/195

### DIFF
--- a/pkg/ipmi/simulator.go
+++ b/pkg/ipmi/simulator.go
@@ -1,22 +1,54 @@
 package ipmi
 
 import (
+	"bytes"
+	"crypto/md5"
+	"crypto/rand"
+	"encoding/binary"
+	"hash/adler32"
 	"net"
+	"sync"
 
 	"github.com/sirupsen/logrus"
 	goipmi "github.com/vmware/goipmi"
+
 	"kubevirt.io/kubevirtbmc/pkg/resourcemanager"
 )
 
+// Command-specific completion codes for IPMI session management.
+// These are in the 0x80-0xBE range per the IPMI spec (section 5.2).
+const (
+	// ActivateSession (22.17): auth failure / invalid AuthCode
+	ccInvalidAuthCode = goipmi.CompletionCode(0x85)
+	// ActivateSession (22.17): invalid session ID
+	ccInvalidSessionID = goipmi.CompletionCode(0x86)
+)
+
+// v15Session holds state for an IPMI v1.5 session challenge-response handshake.
+type v15Session struct {
+	username  string
+	challenge [16]byte
+}
+
+// Simulator provides an IPMI BMC simulator.
 type Simulator struct {
 	ip   string
 	port int
 
 	handler *handler
 	sim     *goipmi.Simulator
+
+	// Authentication credentials
+	username string
+	password string
+
+	// v1.5 session state (temporary session ID -> session state)
+	v15Sessions   map[uint32]*v15Session
+	v15SessionsMu sync.Mutex
 }
 
-func NewSimulator(ip string, port int, resourceManager resourcemanager.ResourceManager) *Simulator {
+// NewSimulator creates a new IPMI simulator with the given credentials.
+func NewSimulator(ip string, port int, resourceManager resourcemanager.ResourceManager, username, password string) *Simulator {
 	return &Simulator{
 		ip:   ip,
 		port: port,
@@ -26,10 +58,14 @@ func NewSimulator(ip string, port int, resourceManager resourcemanager.ResourceM
 			IP:   net.ParseIP(ip).To4(),
 			Port: port,
 		}),
+		username:    username,
+		password:    password,
+		v15Sessions: map[uint32]*v15Session{},
 	}
 }
 
 func (s *Simulator) initialize() {
+	// Override chassis command handlers
 	s.sim.SetHandler(
 		goipmi.NetworkFunctionChassis,
 		goipmi.CommandChassisStatus,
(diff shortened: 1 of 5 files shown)


anishbista60:
Hi @halfcrazy. Thanks for the PR. I am not sure whether this is worth looking into, but from a security point of view I know it is good. However, because these days people are mostly using Redfish than IPMI

halfcrazy:
Yes. Our devices are relatively old, so we mainly rely on IPMI (specifically IPMI v2). That said, fixing the authentication for IPMI v1.5 in advance will let us reuse this part when we later upgrade to IPMI v2.

anishbista60:
> Yes. Our devices are relatively old, so we mainly rely on IPMI (specifically IPMI v2). That said, fixing the authentication for IPMI v1.5 in advance will let us reuse this part when we later upgrade to IPMI v2.

Devices mean are you referring to the baremetal or kubevirt VM ?

halfcrazy:
> > Yes. Our devices are relatively old, so we mainly rely on IPMI (specifically IPMI v2). That said, fixing the authentication for IPMI v1.5 in advance will let us reuse this part when we later upgrade to IPMI v2.
> 
> Devices mean are you referring to the baremetal or kubevirt VM ?

Kubevirt VM. Actually, our deployment tool currently only supports IPMI LANPlus.

anishbista60:
> > > Yes. Our devices are relatively old, so we mainly rely on IPMI (specifically IPMI v2). That said, fixing the authentication for IPMI v1.5 in advance will let us reuse this part when we later upgrade to IPMI v2.
> > 
> > 
> > Devices mean are you referring to the baremetal or kubevirt VM ?
> 
> Kubevirt VM. Actually, our deployment tool currently only supports IPMI LANPlus.

Then no issue. I was thinking how kubevirtbmc gonna work with baremetal it's for k8s env.

anishbista60 (commented):
Hey, Thanks for the PR. Before moving with the review, seems we need to wait for server side authentication as per the comment https://github.com/kubevirtbmc/kubevirtbmc/issues/30#issuecomment-4678529237

halfcrazy:
This fix is implemented based on vmware/goipmi for user authentication, with no reliance on the bougou/go-ipmi implementation. When we later implement LANPlus support, we will migrate to using bougou/go-ipmi.
The function signatures in the section below remain identical; only the internal logic will need modification afterward. Given the upcoming migration, this MR is purely a temporary transitional solution.
```
// NewSimulator creates a new IPMI simulator with the given credentials.
func NewSimulator(ip string, port int, resourceManager resourcemanager.ResourceManager, username, password string) *Simulator
```

pkg/ipmi/simulator.go
```diff
@@ -54,5 +258,5 @@ func (s *Simulator) Run() error {
 
 func (s *Simulator) Stop() {
 	s.sim.Stop()
-	logrus.Info("Redfish emulator gracefully stopped")
+	logrus.Info("IPMI simulator gracefully stopped")
```
anishbista60, line 261: why changes to the Redfish simulator for log ?

anishbista60 (commented):
If I understand the approach correctly,  server validate the client username with the username that is passed as the secret, session will be created, session will be active, validated with authcode and finally server mark the session active on all match. 

Now, one question what if bmc pod gets restarted ?

test/virtbmc-agent/agent_test.go
```diff
@@ -164,6 +195,36 @@ var _ = Describe("Agent e2e", Ordered, func() {
...
+					Username: "baduser",
+					Password: env.Password,
+				})
+				Expect(err).NotTo(HaveOccurred())
+				Expect(out).To(And(
+					ContainSubstring("401"),
+					ContainSubstring("Unauthorized"),
+				))
+			})
 		})
```
anishbista60, line 228: Why redfish related changes ?

pkg/ipmi/simulator.go
halfcrazy: I think the original text is a typo. This file is for ipmi **pkg/ipmi/simulator.go**.

test/virtbmc-agent/agent_test.go
halfcrazy: I added the failed assertions on the fly; only the successful login scenario was covered before.

halfcrazy:
g'g

> If I understand the approach correctly, server validate the client username with the username that is passed as the secret, session will be created, session will be active, validated with authcode and finally server mark the session active on all match.
> 
> Now, one question what if bmc pod gets restarted ?

When the virtbmc pod restarts, all in-memory session state is lost. The client will detect this on its next request and automatically re-establish a new session through the IPMI handshake — no persistence needed.

pkg/ipmi/simulator.go
anishbista60: The naming was wrong in the original file lol :)

anishbista60:
> g'g
> 
> > If I understand the approach correctly, server validate the client username with the username that is passed as the secret, session will be created, session will be active, validated with authcode and finally server mark the session active on all match.
> > Now, one question what if bmc pod gets restarted ?
> 
> When the virtbmc pod restarts, all in-memory session state is lost. The client will detect this on its next request and automatically re-establish a new session through the IPMI handshake — no persistence needed.

yep, Thanks for the clarification.

anishbista60 (approved):
Thanks for the PR. LGTM

/cc @starbops

pkg/ipmi/simulator.go
starbops: my bad lol

starbops (approved):
Works perfectly. Thank you!

starbops:
@mergify backport v0.8

PR #198, opened 2026-06-12
fix: sanitize TAG for Docker image tag compatibility in e2e tests

starbops:
Git branch names with characters like '/' are invalid in Docker image tags. Replace all characters outside [a-zA-Z0-9_.-] with '-' when deriving TAG from VERSION, so kind load docker-image succeeds on any branch name.


anishbista60 (approved):
Thanks for the fix. LGTM

PR #199, opened 2026-06-12
test: resolve kind binary from project-local ./bin/ first

starbops:
Prefer `./bin/kind` installed by `make kind` over whatever is on $PATH. Falls back to the bare name if the local binary is absent, preserving behaviour for environments that haven't run `make kind` yet.


anishbista60 (approved):
Thanks for the enhancement !

PR #181, opened 2026-05-10
feat: Add Toggle to Disable IPMI Support in VirtualMachineBMC CRD

anishbista60:
Tracking issue: #155

PR #200, opened 2026-06-12
fix: update the latest crd for helmchart as well

anishbista60:
- After PR https://github.com/kubevirtbmc/kubevirtbmc/pull/181, there is update in the crd with we forgot to update for the helm chart

This PR fix that replacing old crd to latest one.


starbops (approved):
LGTM, thank you.

PR #209, opened 2026-06-22
docs: update project logo [ci skip]

starbops:
There has been an inconsistency in project logos between the main project and other places. This PR tries to fix that.


starbops:
I will add all the form factors so that they can be reused elsewhere, such as on the documentation site.

Edit: done.

anishbista60 (commented):
Nit: All these form of image really needed ?

starbops:
> Nit: All these form of image really needed ?

Probably not. I was just trying to anticipate different use cases and make the image available in the formats people might need. That said, I'm happy to trim the list down if some of them don't provide much value.

anishbista60 (approved):
Let's keep all of them. Thanks for the PR !

PR #220, opened 2026-06-29
fix(power): prevent power-on from being silently swallowed during VM transitional states

### LINKED ISSUE #219: [BUG] Power operations are unavailable while the virtual machine is in a transitional state
**Describe the bug**

Calling power-on while the virtual machine is shutting down results in a no-op, and the VM remains powered off in the end.

**To Reproduce**

Wait until the virtual machine finishes booting, then run the power off command followed by the power on command.

**Expected behavior**

The virtual machine finishes powering on, or the power-on command for the VM returns an error.
**Screenshots**

- OS: [e.g. SLES 15 SP5]
- Chart version: [e.g. 0.2.3]
- App version: [e.g. v0.5.3]

**Additional context**

halfcrazy:
fix #219 
When a VM is powered off, the VMI may still exist briefly during cleanup. Calling power-on in this window caused KubeVirt to report "VM is already running" (due to the lingering VMI), which the handler previously treated as an idempotent success — returning CommandCompleted (0x00). The VM never actually started, but the client saw success.

Root cause: VMI existence ≠ VM running. During Stopping→Stopped transitions, the VMI cleanup is asynchronous and can outlive the VM's Ready=false state.

Fix: three-layer defense with no dependency on KubeVirt internal error strings.

1. resourcemanager: Try-Then-Verify pattern
 - Call Start/Stop directly, without pre-checking VMI state.
 - On failure, verify VM.Status.Ready to distinguish genuine idempotency from transitional-state failures.
 - Return *ErrRetryable when Ready != desired (transitional state).
 - Return nil when Ready == desired (true idempotent success).

2. handler (IPMI): errors.As(*ErrRetryable) → CompletionCode 0xC0 (Node Busy)
 - 0xC0 per IPMI spec §5.2: "node busy, retry recommended".
 - Zero string matching on KubeVirt error messages.

3. handler (Redfish): errors.As(*ErrRetryable) → HTTP 500 + iLO keyword
 - Based on Ironic's implementation, retries will be triggered when the
   HTTP 5xx status code is returned together with a response body containing
   both iLO and InvalidOperationForSystemState.

4. e2e: verifies power-off → immediate power-on retry loop → VM Running.


starbops (commented):
I haven't checked the code changes, but the PR description concerns me:

```

3. handler (Redfish): errors.As(*ErrRetryable) → HTTP 500 + iLO keyword

   Based on Ironic's implementation, retries will be triggered when the
   HTTP 5xx status code is returned together with a response body containing
   both iLO and InvalidOperationForSystemState.
```

Is there a better way to handle this instead of locking ourselves into vendor-specific code?

halfcrazy:
> I haven't checked the code changes, but the PR description concerns me:
> 
> ```
> 
> 3. handler (Redfish): errors.As(*ErrRetryable) → HTTP 500 + iLO keyword
> 
>    Based on Ironic's implementation, retries will be triggered when the
>    HTTP 5xx status code is returned together with a response body containing
>    both iLO and InvalidOperationForSystemState.
> ```
> 
> Is there a better way to handle this instead of locking ourselves into vendor-specific code?

Sure, 503+retry-after header+ messageId Base.1.13.0.ServiceTemporarilyUnavailable

test/virtbmc-agent/agent_test.go
```diff
@@ -168,6 +168,34 @@ var _ = Describe("Agent e2e", Ordered, func() {
...
+					_, stderr, err := runIPMIInCluster(ctx, config, ns, ipmiReq("power", "on"))
+					if err != nil {
+						sawNodeBusy = true
+						Expect(stderr).To(ContainSubstring("Node busy"),
+							"expected Node Busy retryable error, got stderr=%q", stderr)
+					}
+					return err
+				}, vmPowerStatusTimeout, agentTestInterval).Should(Succeed(),
+					"power on should eventually succeed after VMI cleanup")
+				Expect(sawNodeBusy).To(BeTrue(),
```
halfcrazy, line 211: It seems we need to increase the pod termination‑grace‑period‑seconds for this use‑case.

Its flakey

halfcrazy: Weird, it keeps failing on CI but works fine locally.

halfcrazy:
upstream tracking https://github.com/bougou/go-ipmi/issues/70

test/virtbmc-agent/agent_test.go
halfcrazy: After Stop(), the VMI may still report Ready=true for a brief window (~5-50ms on fast clusters) while runStrategy is already Halted. During this window, Start() fails with 'VM is already running' but the VM is actually about to stop — treating this as idempotent success silently swallows the power-on command.

anishbista60 (commented):
Hi, Thanks alot for the enhancement. I have tested in my local and following are the observation. 

**IPMI**

1. Response time is noticeably slow, each power command takes almost 2 seconds, whereas on the upstream (unfixed) code it completes within milliseconds. 

2. When powering on via IPMI and immediately issuing another power-on, it fails with: `Set Chassis Power Control to Up/On failed: Node busy`. Is this expected behavior?

3. When powering off and then immediately issuing another power-off, there is no response/output at all. Is this expected behavior?

**Redfish**

1. When powering off via Redfish and immediately powering on, it returns: `"SystemState - Power transition in progress. Retry later."` This  looks good. 

2. When powering on and immediately powering off, the VM get turn off, without returning a message indicating the system is in a transitional state like above. Is this expected behavior?

halfcrazy:
> Hi, Thanks alot for the enhancement. I have tested in my local and following are the observation.
> 
> **IPMI**
> 
> 1. Response time is noticeably slow, each power command takes almost 2 seconds, whereas on the upstream (unfixed) code it completes within milliseconds.

I suspect the slow response comes from the virt-api in your environment. There isn’t much room left for optimization on the code side.

| Scenario | Upstream (legacy) virtbmc side | This branch virtbmc side |
| ---- | ---- | ---- |
| Forward: Off → On | GET VMI + Start = 2 RTTs | Start = 1 RTT |
| Forward: On → Off | GET VMI + Stop = 2 RTTs | Stop = 1 RTT |
| Idempotent steady state: Start when already powered on | GET VMI returns immediately = 1 RTT | Start (failed) + GET VM = 2 RTTs |
| Idempotent steady state: Stop when already powered off | GET VMI returns immediately = 1 RTT | Stop (failed) + GET VM = 2 RTTs |
| Transition state: Power on during shutdown cleanup | GET VMI (exists) misjudged as success = 1 RTT | Start (failed) + GET VM = 2 RTTs |

> 2. When powering on via IPMI and immediately issuing another power-on, it fails with: `Set Chassis Power Control to Up/On failed: Node busy`. Is this expected behavior?

Nope, I think this should return ok, this operation is idempotent. I’ll look into this. Could you share the YAML of the VM you used for testing? Also, what `RunStrategy` is configured for this VM?

> 3. When powering off and then immediately issuing another power-off, there is no response/output at all. Is this expected behavior?

Yes, this operation is idempotent.

> 
> **Redfish**
> 
> 1. When powering off via Redfish and immediately powering on, it returns: `"SystemState - Power transition in progress. Retry later."` This  looks good.

Yes

> 2. When powering on and immediately powering off, the VM get turn off, without returning a message indicating the system is in a transitional state like above. Is this expected behavior?

Yes

anishbista60:
> > 2. When powering on via IPMI and immediately issuing another power-on, it fails with: `Set Chassis Power Control to Up/On failed: Node busy`. Is this expected behavior?
> 
> Nope, I think this should return ok, this operation is idempotent. I’ll look into this. Could you share the YAML of the VM you used for testing? Also, what `RunStrategy` is configured for this VM?
> 

I copy paste the VM manifest from our docs. https://docs.kubevirtbmc.io/getting-started/#create-your-first-virtual-bmc

anishbista60:
> I suspect the slow response comes from the virt-api in your environment. There isn’t much room left for optimization on the code side.
> 

I have tested in the same environment by switching the main branch and your branch.  Not sure virt-api cause this.

halfcrazy:
> > I suspect the slow response comes from the virt-api in your environment. There isn’t much room left for optimization on the code side.
> 
> I have tested in the same environment by switching the main branch and your branch. Not sure virt-api cause this.

Confirmed. The issue is caused by go-ipmi upstream dual-stack handler. Will fix in upstream.

```
/ # time ipmitool -I lanplus -H 10.96.3.7 -U admin -P password  chassis power status
Chassis Power is on
real	0m 0.00s
user	0m 0.00s
sys	0m 0.00s
/ # time ipmitool -I lan -H 10.96.3.7 -U admin -P password  chassis power status
Chassis Power is on
real	0m 2.01s
user	0m 0.00s
sys	0m 0.00s
```

halfcrazy:
xref https://github.com/bougou/go-ipmi/pull/75

halfcrazy:
> xref [bougou/go-ipmi#75](https://github.com/bougou/go-ipmi/pull/75)

Should work as before. Hold until upstream merged.

replace github.com/bougou/go-ipmi => github.com/halfcrazy/go-ipmi v0.0.0-20260718035953-210ae29f9b8d
```
/ # time ipmitool -I lan -H 10.96.22.140 -U admin -P password chassis power status
Chassis Power is on
real	0m 0.00s
user	0m 0.00s
sys	0m 0.00s
/ # time ipmitool -I lanplus -H 10.96.22.140 -U admin -P password chassis power status
Chassis Power is on
real	0m 0.00s
user	0m 0.00s
sys	0m 0.00s
```

anishbista60:
> > I suspect the slow response comes from the virt-api in your environment. There isn’t much room left for optimization on the code side.
> 
> I have tested in the same environment by switching the main branch and your branch. Not sure virt-api cause this.

Cool. Let's wait.

halfcrazy:
> > xref [bougou/go-ipmi#75](https://github.com/bougou/go-ipmi/pull/75)
> 
> Should work as before. Hold until upstream merged.
> 
> replace github.com/bougou/go-ipmi => github.com/halfcrazy/go-ipmi v0.0.0-20260718035953-210ae29f9b8d
> 
> ```
> / # time ipmitool -I lan -H 10.96.22.140 -U admin -P password chassis power status
> Chassis Power is on
> real	0m 0.00s
> user	0m 0.00s
> sys	0m 0.00s
> / # time ipmitool -I lanplus -H 10.96.22.140 -U admin -P password chassis power status
> Chassis Power is on
> real	0m 0.00s
> user	0m 0.00s
> sys	0m 0.00s
> ```

The upstream merge is done. Bump version in this branch.

pkg/resourcemanager/virtualmachine_resourcemanager.go
```diff
@@ -264,74 +263,219 @@ func (m *VirtualMachineResourceManager) GetPowerStatus() (bool, error) {
...
+	// starting. Repeated power-on in this window is idempotent.
+	if rsErr == nil && rs == kubevirtv1.RunStrategyAlways {
+		return nil
+	}
+	// Manual / RerunOnFailure: StartRequest may already be consumed by
+	// virt-controller while the VMI is still starting (Ready=false).
+	// A non-final VMI with no pending stop means power-on is already
+	// underway — idempotent. Final (Failed/Succeeded) VMI is not treated
+	// as success: RerunOnFailure rejects start-from-failed, and Manual
+	// needs an explicit restart path.
```
anishbista60: one last nit: Decrease the length of the code comment / remove the code comment where it is not necessary from dev perspective across the whole PR. 

I will get some time to test it again during weekend and then I will approve it. Thanks

anishbista60:
> **IPMI**
> 
> 1. Response time is noticeably slow, each power command takes almost 2 seconds, whereas on the upstream (unfixed) code it completes within milliseconds.
> 2. When powering on via IPMI and immediately issuing another power-on, it fails with: `Set Chassis Power Control to Up/On failed: Node busy`. Is this expected behavior?



Cool, IPMI Part works perfectly 

There were no issue with Redfish in  last testing and even no issue at now. 

https://github.com/user-attachments/assets/68bd7aed-2092-4b7b-8548-e498e45e6b7b

anishbista60 (approved):
LGTM. Thanks for the enhancement 

/cc @starbops

pkg/redfish/api_service.go
```diff
@@ -59861,8 +59861,28 @@ func (s *APIService) RedfishV1SystemsComputerSystemIdActionsComputerSystemResetP
...
+					Message: "iLO: InvalidOperationForSystemState - Power transition in progress. Retry later.",
+					MessageExtendedInfo: []server.MessageV120Message{
+						{
+							MessageId: "iLO.2.13.InvalidOperationForSystemState",
+							Message:   "InvalidOperationForSystemState: Power transition in progress",
+						},
+					},
+				},
+			}), nil
+		}
```
starbops, line 59882: Is it possible to decouple from vendor specific terms?

halfcrazy: > Is it possible to decouple from vendor specific terms?

I haven’t come up with a better approach. The Node Busy completion code for IPMI is a general mechanism. There does not appear to be a unified specification for the Redfish portion, and my current implementation is inferred based on Ironic.

starbops: > > Is it possible to decouple from vendor specific terms?
> 
> I haven’t come up with a better approach. The Node Busy completion code for IPMI is a general mechanism. There does not appear to be a unified specification for the Redfish portion, and my current implementation is inferred based on Ironic.

Do you think it would be viable to propose a corresponding change to Ironic to support multiple keywords, if we decide to coin our own? That way, we'd simply become another implementation alongside HPE iLO, rather than effectively impersonating another vendor by using its vendor-specific term.

That said, I'm fine with the current approach as long as it works with Ironic out of the box. I do think this is worth documenting somewhere, though, so users understand why we made this choice. We can always revisit it and pursue a more proper solution in the future. WDYT?

halfcrazy: > > > Is it possible to decouple from vendor specific terms?
> > 
> > 
> > I haven’t come up with a better approach. The Node Busy completion code for IPMI is a general mechanism. There does not appear to be a unified specification for the Redfish portion, and my current implementation is inferred based on Ironic.
> 
> Do you think it would be viable to propose a corresponding change to Ironic to support multiple keywords, if we decide to coin our own? That way, we'd simply become another implementation alongside HPE iLO, rather than effectively impersonating another vendor by using its vendor-specific term.
> 
> That said, I'm fine with the current approach as long as it works with Ironic out of the box. I do think this is worth documenting somewhere, though, so users understand why we made this choice. We can always revisit it and pursue a more proper solution in the future. WDYT?

If external exposure is considered, I tend to file an issue with KubeVirt to make its power operation interface eventually consistent. Instead of notifying callers to retry via specific return values, another approach is to implement polling inside the program and return only after the operation succeeds. However, this can easily lead to timeouts for ipmitool/Redfish calls. 
I can add documentation to the docs repository describing our current handling behavior for intermediate states (returning specific responses).

Mode 1: Ironic-compatible “retryable error” mode // current method
When the VM is in a transitional/asynchronous state, virtbmc should return a specific retryable error instead of trying to wait for completion.

IPMI: return a completion code like Node Busy so clients (e.g., Ironic/ipmitool) can retry.
Redfish: respond with an HTTP error (e.g., 500) and a message/body pattern that makes sushy/Ironic trigger its retry logic. This mode is optimized for clients that already implement “retry on transient/busy/state-transition” behavior.

Mode 2: Tinkerbell/Foreman-compatible “server wait” mode // just for discussion and if there is a usecase we can have a try
When the VM is in a transitional state, virtbmc should internally poll until the operation is actually successful, and only then return success to the client.

For power on/off/cycle, virtbmc keeps checking VM/VMI state until it reaches the final desired state (using your existing “real success” criteria).
This mode is optimized for clients that do not reliably retry based on HTTP/IPMI error codes.
In short:

Mode 1 returns “retryable errors” and relies on client retries.
Mode 2 waits inside virtbmc and returns only when the state is truly settled.

starbops: That's also a viable direction—provided KubeVirt considers it a gap that should be addressed.

That said, I'm fine with the current approach. Let's document the rationale behind it.

Thank you!

starbops (approved):

PR #221, opened 2026-06-30
added feature to support additional network on virtbmc pod

### LINKED ISSUE #150: [FEATURE] Enabling virtual BMCs to connect to networks other than the default pod network.

The claim "out-of-band management of KubeVirt virtual machines" becomes less appealing when all operations are still performed on the default pod network. Although users can expose BMC access endpoints to the external via Ingress or Gateway APIs (not yet explored!), this may not always be the case. Because BMCs are usually vulnerable and have high privileges (direct control over hardware), they are typically resident in an isolated network. Such a network is unlikely to be reachable from the other, but the provisioning tools.

A typical use case would be to run both the to-be-provisioned virtual machines and their virtual BMCs in separate networks, and to have the provisioning stack reach both networks to conduct out-of-band management and provisioning tasks, respectively.

Make VirtualMachineBMC CRD supports specifying the network (NetworkAttachmentDefinition) that the instantiated virtual BMC pods should attach to. This would require [Multus](https://github.com/k8snetworkplumbingwg/multus-cni) as a dependency.

With CNI such as Kube-OVN, pods can attach to user-defined Subnets, enabling even versatile use cases. For instance, with properly configured Kube-OVN Subnets, the user creates VirtualMachine and VirtualMachineBMC resources under the `production` namespace, [the instantiated virtual BMC pod will be automatically attached to the Subnet that associated with the namespace](https://kubeovn.github.io/docs/v1.15.x/en/guide/subnet/#create-pod-in-the-subnet). On the other hand, [VirtualMachine can attach to a different Subnet through properly configured NetworkAttachmentDefinition](https://kubeovn.github.io/docs/v1.15.x/en/advance/multi-nic/?h=networkattachment#usage).

**Additional context**

N/A

devenkulkarni:
This implementation adds a single additional network via spec.networkRef, preserving the default Pod network. The controller recreates the virtBMC Pod when networkRef changes because Pod network attachments are immutable.

Fixes #150

### DIFF
--- a/internal/controller/virtualmachinebmc/controller.go
+++ b/internal/controller/virtualmachinebmc/controller.go
@@ -507,7 +507,7 @@ func podNetworkRef(pod *corev1.Pod) string {
 // reconcileNetworkChange detects a mismatch between the spec's NetworkRef and
 // the annotation stamped on the existing pod. Since pod networking is immutable,
 // the pod must be recreated when the network attachment changes.
-func (r *VirtualMachineBMCReconciler) reconcileNetworkChange(ctx context.Context, virtualMachineBMC *bmcv1.VirtualMachineBMC) (requeue bool, err error) {
+func (r *VirtualMachineBMCReconciler) reconcileNetworkChange(ctx context.Context, virtualMachineBMC *bmcv1.VirtualMachineBMC) (deleted bool, err error) {
 
 	log := log.FromContext(ctx)
 
@@ -611,7 +611,7 @@ func (r *VirtualMachineBMCReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return ctrl.Result{}, err
 	}
 	if deleted {
-		return ctrl.Result{Requeue: true}, nil
+		return ctrl.Result{}, nil
 	}
 
 	// Prepare the virtBMC Pod
--- a/internal/controller/virtualmachinebmc/controller.go
+++ b/internal/controller/virtualmachinebmc/controller.go
@@ -140,17 +140,23 @@ func (r *VirtualMachineBMCReconciler) constructPodFromVirtualMachineBMC(virtualM
 		secretName = virtualMachineBMC.Spec.AuthSecretRef.Name
 	}
 
+	annotations := map[string]string{
+		EnableIPMIAnnotation: strconv.FormatBool(specIPMIEnabled(&virtualMachineBMC.Spec)),
+	}
+
+	if virtualMachineBMC.Spec.NetworkRef != "" {
+		annotations[MultusNetworksAnnotation] = virtualMachineBMC.Spec.NetworkRef
+	}
+
 	pod := &corev1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
 			Labels: map[string]string{
 				VirtualMachineBMCNameLabel: virtualMachineBMC.Name,
 				VMNameLabel:                virtualMachineBMC.Spec.VirtualMachineRef.Name,
 			},
-			Annotations: map[string]string{
-				EnableIPMIAnnotation: strconv.FormatBool(specIPMIEnabled(&virtualMachineBMC.Spec)),
-			},
-			Name:      name,
-			Namespace: virtualMachineBMC.Namespace,
+			Annotations: annotations,
+			Name:        name,
+			Namespace:   virtualMachineBMC.Namespace,
 		},
 		Spec: corev1.PodSpec{
 			ServiceAccountName: serviceAccountName,
@@ -486,6 +492,52 @@ func (r *VirtualMachineBMCReconciler) reconcileIPMIChange(ctx context.Context, v
 	return true, nil
 }
 
+func specNetworkRef(spec *bmcv1.VirtualMachineBMCSpec) string {
+	return spec.NetworkRef
+}
+
+func podNetworkRef(pod *corev1.Pod) string {
+	if pod == nil {
+		return ""
+	}
+
+	return pod.Annotations[MultusNetworksAnnotation]
+}
+
+// reconcileNetworkChange detects a mismatch between the spec's NetworkRef and
+// the annotation stamped on the existing pod. Since pod networking is immutable,
+// the pod must be recreated when the network attachment changes.
+func (r *VirtualMachineBMCReconciler) reconcileNetworkChange(ctx context.Context, virtualMachineBMC *bmcv1.VirtualMachineBMC) (requeue bool, err error) {
+
+	log := log.FromContext(ctx)
+
+	pod, err := r.getVirtBMCPod(ctx, virtualMachineBMC)
+	if err != nil {
+		return false, err
+	}
+
+	if pod == nil {
+		return false, nil
+	}
+
+	currentNetworkRef := podNetworkRef(pod)
+	desiredNetworkRef := specNetworkRef(&virtualMachineBMC.Spec)
+
+	if currentNetworkRef == desiredNetworkRef {
+		return false, nil
+	}
+
+	log.Info("networkRef changed, replacing pod",
+		"currentNetworkRef", currentNetworkRef,
+		"desiredNetworkRef", desiredNetworkRef)
+
+	if err := r.deleteVirtBMCPod(ctx, virtualMachineBMC); err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
+
 //+kubebuilder:rbac:groups=kubevirt.io,resources=virtualmachines,verbs=get;list;watch;update;patch
 //+kubebuilder:rbac:groups=kubevirt.io,resources=virtualmachineinstances,verbs=get;list;watch;delete
 //+kubebuilder:rbac:groups=bmc.kubevirt.io,resources=virtualmachinebmcs,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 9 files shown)


devenkulkarni:
I have not added/updated anything in *_test.go yet cause I might need some help with that, although I have tested this locally in a kind cluster.

anishbista60:
> I have not added/updated anything in *_test.go yet cause I might need some help with that, although I have tested this locally in a kind cluster.


Thanks for the PR. Leave this PR with me. I will take care by tomorrow evening. I am not able see the changes properly via phone lol 🥲

internal/controller/virtualmachinebmc/controller.go
```diff
@@ -486,6 +492,52 @@ func (r *VirtualMachineBMCReconciler) reconcileIPMIChange(ctx context.Context, v
...
+		return ""
+	}
+
+	return pod.Annotations[MultusNetworksAnnotation]
+}
+
+// reconcileNetworkChange detects a mismatch between the spec's NetworkRef and
+// the annotation stamped on the existing pod. Since pod networking is immutable,
+// the pod must be recreated when the network attachment changes.
+func (r *VirtualMachineBMCReconciler) reconcileNetworkChange(ctx context.Context, virtualMachineBMC *bmcv1.VirtualMachineBMC) (requeue bool, err error) {
```
anishbista60:
```suggestion
func (r *VirtualMachineBMCReconciler) reconcileNetworkChange(ctx context.Context, virtualMachineBMC *bmcv1.VirtualMachineBMC) (deleted bool, err error) {
```

internal/controller/virtualmachinebmc/controller.go
```diff
@@ -554,6 +606,14 @@ func (r *VirtualMachineBMCReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return ctrl.Result{}, err
 	}
 
+	deleted, err = r.reconcileNetworkChange(ctx, &virtualMachineBMC)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if deleted {
+		return ctrl.Result{Requeue: true}, nil
```
anishbista60:
```suggestion
		return ctrl.Result{}, nil
```

Since, controller is setup with `Owns(&corev1.Pod{})`, next reconcile loop will recreate the pod automatically. 

Same changes is also required for `reconcileIPMIChange` that I missed at that time :)

anishbista60 (commented):
Thanks for PR. We need to update the controller_test.go file and further e2e test for the controller at https://github.com/kubevirtbmc/kubevirtbmc/blob/main/test/virtbmc-controller/controller_test.go

anishbista60:
Did you verified this part  https://github.com/kubevirtbmc/kubevirtbmc/issues/150#issuecomment-4677872803 ?

devenkulkarni:
> Did you verified this part [#150 (comment)](https://github.com/kubevirtbmc/kubevirtbmc/issues/150#issuecomment-4677872803) ?

No, because here - https://github.com/kubevirtbmc/kubevirtbmc/issues/150#issuecomment-4677967684 @starbops mentioned on adding additional network instead of replacing default pod network.

devenkulkarni:
> Thanks for PR. We need to update the controller_test.go file and further e2e test for the controller at https://github.com/kubevirtbmc/kubevirtbmc/blob/main/test/virtbmc-controller/controller_test.go

I might need some help with this to understand the flow.

internal/controller/virtualmachinebmc/controller.go
devenkulkarni: Okay I will correct the suggested changes for the changes I made and I think we can take IPMI changes later in separate PR right?

anishbista60: yes, you can do it in next PR.

starbops:
> > Did you verified this part [#150 (comment)](https://github.com/kubevirtbmc/kubevirtbmc/issues/150#issuecomment-4677872803) ?
> 
> No, because here - [#150 (comment)](https://github.com/kubevirtbmc/kubevirtbmc/issues/150#issuecomment-4677967684) @starbops mentioned on adding additional network instead of replacing default pod network.

We don't need to worry that because the default is the pod network. The healthiness/readiness probe still works.

halfcrazy:
> > Thanks for PR. We need to update the controller_test.go file and further e2e test for the controller at https://github.com/kubevirtbmc/kubevirtbmc/blob/main/test/virtbmc-controller/controller_test.go
> 
> I might need some help with this to understand the flow.

You can take a look at  https://github.com/halfcrazy/kubevirtbmc/commit/9613dc80290d3d124d189da47d8e8a3cb1f4517a

devenkulkarni:
> > > Thanks for PR. We need to update the controller_test.go file and further e2e test for the controller at https://github.com/kubevirtbmc/kubevirtbmc/blob/main/test/virtbmc-controller/controller_test.go
> > 
> > 
> > I might need some help with this to understand the flow.
> 
> You can take a look at [halfcrazy@9613dc8](https://github.com/halfcrazy/kubevirtbmc/commit/9613dc80290d3d124d189da47d8e8a3cb1f4517a)

I added tests in controller_test.go but your commit helps as its much more than what I added. I am learning these with the help of AI. Can you explain these tests to me in a simpler way? 

TBH, I just understood the concept of why we need these tests, but I am still learning and understanding the code.

halfcrazy:
> > > > Thanks for PR. We need to update the controller_test.go file and further e2e test for the controller at https://github.com/kubevirtbmc/kubevirtbmc/blob/main/test/virtbmc-controller/controller_test.go
> > > 
> > > 
> > > I might need some help with this to understand the flow.
> > 
> > 
> > You can take a look at [halfcrazy@9613dc8](https://github.com/halfcrazy/kubevirtbmc/commit/9613dc80290d3d124d189da47d8e8a3cb1f4517a)
> 
> I added tests in controller_test.go but your commit helps as its much more than what I added. I am learning these with the help of AI. Can you explain these tests to me in a simpler way?
> 
> TBH, I just understood the concept of why we need these tests, but I am still learning and understanding the code.

Testing for this project falls into two categories: unit tests and integration tests.
The unit test suite includes tests that mock the apiserver using envtest.
All integration tests reside under the /test/ directory and run against a real Kubernetes cluster provisioned via Kind.

As infrastructure components, these require minimal attention. Most of the work revolves around preloading data into the apiserver and validating expected outcomes. However, envtest only provides storage capabilities and cannot replicate state transitions that occur in a real Kubernetes environment, such as pod scheduling and runtime behaviors.

anishbista60 (dismissed):
Till now it is good to go. Please add  e2e test then we will have another round of review.

anishbista60:
@devenkulkarni Please fix the lint issue locally and to verify run `make lint`

test/virtbmc-controller/controller_test.go
```diff
@@ -332,5 +332,76 @@ var _ = Describe("KubeVirtBMC controller manager", Ordered, func() {
...
+					}
+					if pod.UID == originalUID {
+						return false
+					}
+					val, ok := pod.Annotations[util.MultusNetworksAnnotation]
+					return ok && val == util.E2EMultusNetworkName
+				}, timeout, interval).Should(BeTrue(), "Pod should be recreated with Multus network annotation")
+
+				// NOTE: we do NOT require the pod to reach Running state here.
+				// The delegate CNI (bridge) may fail in Kind depending on
```
anishbista60, line 367: May fail or acutally failing ? I feel we need to verify it. At the end of the day we need pod running.

test/util/helper.go
```diff
@@ -15,6 +15,49 @@ const (
...
 	kubeVirtOperatorURLFmt = "https://github.com/kubevirt/kubevirt/releases/download/%s/kubevirt-operator.yaml"
 	kubeVirtCRURLFmt       = "https://github.com/kubevirt/kubevirt/releases/download/%s/kubevirt-cr.yaml"
+
+	// NADCRDYAML is the NetworkAttachmentDefinition CRD definition.
+	// We apply only the CRD (not the full Multus daemonset) so the
+	// k8s.v1.cni.cncf.io/networks pod annotation is pure metadata —
+	// no actual CNI plugin runs, the pod starts normally with kindnet.
+	NADCRDYAML = `apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
```
anishbista60, line 25: where we are getting this crd from ?

starbops: @devenkulkarni I don't see an issue with installing Multus here. We already follow the same pattern for KubeVirt and cert-manager, and given that this is an E2E test, the additional installation step seems reasonable.

anishbista60: I wish if we can get this CRD using URL, then we could add the url as variable and write a small function to apply that in the cluster rather than hardcording it.

anishbista60: @devenkulkarni Do you need any help ?

anishbista60: I would love to see this PR in `v0.9.0` release.

halfcrazy: You can take a look at https://github.com/halfcrazy/kubevirtbmc/commit/937586f01cc915c8b3b23ce5393b6ae15deeab49, which adds a real multus+bridge case.

PR #225, opened 2026-07-01
fix(ipmi): swap PowerCycle and HardReset method mappings

halfcrazy:
Per IPMI spec §28.3, Power Cycle (0x02) turns the system off then back on — cutting power entirely — which is the most disruptive reset. Hard Reset (0x03) asserts the system reset line without power cycling, making it strictly less brutal.

The previous mapping had these backwards:
  ControlPowerCycle    → PowerCycle()      (graceful restart)
  ControlPowerHardReset → ForcePowerCycle() (force restart)

Swap them so the more brutal operation maps to the force variant:
  ControlPowerCycle    → ForcePowerCycle() (GracePeriodSeconds=0)
  ControlPowerHardReset → PowerCycle()      (graceful restart)

fix https://github.com/kubevirtbmc/docs/pull/14#pullrequestreview-4608132557

### DIFF
--- a/pkg/ipmi/handler.go
+++ b/pkg/ipmi/handler.go
@@ -38,13 +38,16 @@ func (h *handler) chassisControlHandler(m *goipmi.Message) goipmi.Response {
 		logrus.Info("power on")
 		err = h.rm.PowerOn()
 	case goipmi.ControlPowerCycle:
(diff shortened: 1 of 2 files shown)


starbops (approved):
LGTM, thanks for the quick fix!

anishbista60 (approved):
LGTM. Thanks !

Issue #232, opened 2026-07-02
Call for Adopters - help us update ADOPTERS.md


anishbista60:
We have an [ADOPTERS.md](https://github.com/kubevirtbmc/kubevirtbmc/blob/main/ADOPTERS.md) file.

If you're using KubeVirtBMC, please comment below and let us know, we'll help get you added to the list, or feel free to open a PR yourself following the instructions in the file.

Thanks for supporting the project!

PR #233, opened 2026-07-02
Revise adopters table and add CESTC entry

halfcrazy:
Updated the adopters table format and added CESTC as an adopter.

related issue: #232


anishbista60 (approved):
LGTM. Thank you so much !

PR #234, opened 2026-07-02
feat(ipmi): migrate to bougou/go-ipmi for RMCP+ LANPLUS support

### LINKED ISSUE #30: [FEATURE] RMCP+ support

halfcrazy:
Replace the custom IPMI v1.5 session implementation (github.com/vmware/goipmi) with bougou/go-ipmi's full RMCP+/IPMI v2.0 LANPLUS stack.

Key changes:

- Switch IPMI library from vmware/goipmi to bougou/go-ipmi, which provides typed IPMI message codecs, RMCP+ framing, RAKP authentication, and built-in session management per the IPMI v2.0/RMCP+ spec.

- Implement hal.ChassisHAL via vmChassis (handler.go), translating each chassis control action specified in Table 28-3 to the corresponding KubeVirt ResourceManager API:
    Power Down (0x00)        -> PowerOff()        (graceful stop)
    Power Up (0x01)          -> PowerOn()
    Power Cycle (0x02)       -> PowerCycle()
    Hard Reset (0x03)        -> ForcePowerCycle() (forceful, grace=0s)
    ACPI Soft Shutdown (0x05)-> PowerOff()        (graceful stop)

- Use go-ipmi's typed BootDeviceSelector (spec Table 28-6) for Set System Boot Options, replacing ad-hoc byte-level parsing. The boot flags are not mirrored in memory — GetBootFlags/GetBootInfoAcknowledge consistently return ErrNotSupported (-> 0x80, per §28.13), as a pod restart would lose any in-memory copy and produce stale state. The authoritative persistent boot order resides in the VM spec.

- Wire go-ipmi's server.Server with a HandlerRegistry (App, Session, Chassis handlers) over a udptransport.Conn, eliminating the previous manual v1.5 challenge-response handshake (v15Session, GetSessionChallenge, ActivateSession). go-ipmi's built-in RMCP+/RAKP handles authentication, integrity, and confidentiality natively.

- Separate lifecycle from business logic: simulator.go owns bind/serve/stop and BMC device-id construction; handler.go contains only the KubeVirt- specific chassis HAL implementation.

- Update unit tests and virtbmc-agent integration tests to exercise the new HAL-based design.

The go-ipmi dependency is pinned to a post-release commit (bougou/go-ipmi v0.8.4-0.20260701115353-093eaccbe633) that includes the latest typed HAL and server infrastructure.

https://github.com/kubevirtbmc/docs/pull/19
fix #30


anishbista60 (commented):
Thanks for the PR !

pkg/ipmi/handler.go
```diff
@@ -1,124 +1,167 @@
...
 }
 
-func (h *handler) chassisControlHandler(m *goipmi.Message) goipmi.Response {
-	r := &goipmi.ChassisControlRequest{}
-	if err := m.Request(r); err != nil {
-		return err
-	}
+func (c vmChassis) PowerState(ctx context.Context) (bool, error) {
+	return c.rm.GetPowerStatus()
+}
```
anishbista60, line 65: I don't see this function being used somewhere to check the status of VM.

pkg/ipmi/handler.go
```diff
@@ -1,124 +1,167 @@
...
-		// Immediate power down (no OS notification), per IPMI spec §28.3.
-		logrus.Info("force power off")
-		err = h.rm.ForcePowerOff()
-	case goipmi.ControlPowerAcpiSoft:
-		// ACPI graceful shutdown, per IPMI spec §28.3.
-		logrus.Info("power off")
-		err = h.rm.PowerOff()
-	case goipmi.ControlPowerUp:
+func (c vmChassis) SetPower(ctx context.Context, on bool) error {
+	if on {
```
anishbista60, line 68: I am not able to understand that How this `on` boolean able to distinguish whether VM is on or off ?

pkg/ipmi/handler.go
```diff
@@ -1,124 +1,167 @@
...
-		return &goipmi.ChassisControlResponse{
-			CompletionCode: goipmi.ErrInvalidState,
-		}
+		return c.rm.PowerOn()
 	}
+	// Immediate power down (no OS notification), per IPMI spec §28.3
+	// ControlPowerDown (0x00). Contrast with WarmReset (0x05 ACPI Soft
+	// Shutdown), which is a graceful stop.
+	logrus.Info("force power off")
+	return c.rm.ForcePowerOff()
```
anishbista60, line 76: Do we really need to prefer force off ?

pkg/ipmi/handler.go
```diff
@@ -1,124 +1,167 @@
...
 
-func (h *handler) chassisStatusHandler(*goipmi.Message) goipmi.Response {
-	logrus.Info("power status")
+// ColdReset maps IPMI Hard Reset (0x03) to a graceful KubeVirt restart. Per
+// IPMI spec §28.3, Hard Reset asserts the system reset line without power
+// cycling, making it less disruptive than a full Power Cycle (0x02).
+func (c vmChassis) ColdReset(ctx context.Context) error {
+	logrus.Info("hard reset (graceful restart)")
+	return c.rm.PowerCycle()
+}
```
anishbista60, line 95: It should be the force power cycle right ?

pkg/ipmi/handler.go
```diff
@@ -1,124 +1,167 @@
...
-	}
+// GetBootFlags always returns ErrNotSupported (→ 0x80). The virtual BMC does
+// not persist boot flags: a pod restart would lose any in-memory copy and
+// return values inconsistent with the actual VM spec bootOrder. Reporting
+// "parameter not supported" consistently is preferable to returning stale or
+// divergent state. The persistent boot order lives in the VM spec, written via
+// SetBootFlags → rm.SetBootDevice.
+func (vmChassis) GetBootFlags(context.Context) (*ipmi.BootOptionParam_BootFlags, error) {
+	return nil, hal.ErrNotSupported
+}
```
anishbista60, line 134: Worth a line explaining boot device is set-only via IPMI  in docs  and to check current boot order, look at the VM spec directly (or via Redfish, if that path supports GET).

anishbista60 (commented):
`soft` and `reset` command in the docs  are missing https://github.com/kubevirtbmc/docs/pull/19

pkg/ipmi/handler.go
halfcrazy: Called by go-ipmi's Get Chassis Status handler

https://github.com/bougou/go-ipmi/blob/093eaccbe63326d6566a63bd67d878f9470c2dee/pkg/handlers/chassis.go#L41-L67

pkg/ipmi/handler.go
halfcrazy: It's the desired state, if `on` is not straightforward, maybe renamed to `powerOn`?

pkg/ipmi/handler.go
halfcrazy: Yes, equal power down.

https://github.com/bougou/go-ipmi/blob/093eaccbe63326d6566a63bd67d878f9470c2dee/pkg/handlers/chassis.go#L87-L90
````
	case chassis.ChassisControlPowerDown:
		return nil, codeFromErr(ch.SetPower(ctx, false)), nil
	case chassis.ChassisControlPowerUp:
		return nil, codeFromErr(ch.SetPower(ctx, true)), nil
```

pkg/ipmi/handler.go
anishbista60: Thanks for reference.

pkg/ipmi/handler.go
anishbista60: It's fine. No need to rename.

pkg/ipmi/handler.go
halfcrazy: Good call-out. I looked into whether Redfish GET could serve as the authoritative read path here, and it turns out it has the same fundamental problem — the boot info returned by Redfish `GET /Systems/{id}` also comes from in-memory state in the `ComputerSystemAdapter`, not from the VM spec.

**How Redfish boot GET currently works:**

1. `ComputerSystemAdapter` is initialized with hardcoded defaults (`BootSourceOverrideEnabled: DISABLED`, `BootSourceOverrideTarget: HDD`) — regardless of the actual VM spec bootOrder.
2. After a `SetBootDevice()` call (via Redfish PATCH or IPMI), the in-memory adapter is patched to reflect the last-set override.
3. On pod restart, the in-memory adapter resets to defaults — the GET response no longer matches the actual VM spec.

So Redfish GET for boot device suffers from the same statelessness issue. The difference is only in how the two protocols handle it:

| | IPMI `GetBootFlags` | Redfish `GET /Systems/{id}` Boot |
|---|---|---|
| Response | `ErrNotSupported` (0x80 per spec §28.13) | `Boot` object from in-memory adapter |
| After pod restart | Still 0x80 — consistent | Resets to defaults — **stale/misleading** |

**Why I think keeping `ErrNotSupported` on the IPMI side is the right call:**

- IPMI spec §28.13 explicitly allows returning `CodeBootParamNotSupported` (0x80) when the BMC does not support reading a boot parameter — this is fully spec-compliant.
- Returning an explicit error code is safer than returning data that may silently be wrong after a pod restart.
- The authoritative, persistent boot order lives in the VM spec (`spec.template.spec.domain.devices.disks[*].bootOrder` / `interfaces[*].bootOrder`), written via `SetBootDevice` → JSON Patch. Neither protocol currently has a GET path that reads back from that source.

**Could we add a real GET someday?** Yes — the clean approach would be to persist "last-set boot device" in a VM annotation or the BMC CR status so it survives pod restarts. But that is a separate feature. For now, I think consistency is better: both IPMI and Redfish are set-only for boot, and we should doc that clearly.

pkg/ipmi/handler.go
halfcrazy: No, from the discussion in https://github.com/kubevirtbmc/kubevirtbmc/pull/225
```
Swap them so the more brutal operation maps to the force variant:
ControlPowerCycle → ForcePowerCycle() (GracePeriodSeconds=0)
ControlPowerHardReset → PowerCycle() (graceful restart)
```

pkg/ipmi/handler.go
halfcrazy: I’ve added explanations in the document.

pkg/ipmi/handler.go
starbops: We can rename the internal functions to make them less confusing. As long as the interface remains consistent, that should be fine.

halfcrazy:
Our internal implementation relies on the Product Serial field returned by fru list, so I’ve included this field in the update. It returns the UID of the virtual machine.

```
ipmitool -I lanplus -U "$USERNAME" -P "$PASSWORD" -H "$HOSTNAME" fru list
```

**Output:**
```
FRU Device Description : Builtin FRU Device (ID 0)
 Product Manufacturer  : KubeVirt
 Product Name          : KubeVirtBMC
 Product Version       : <build-version>
 Product Serial        : <vm-metadata-uid>
```

pkg/ipmi/fru.go
```diff
@@ -0,0 +1,228 @@
+package ipmi
+
+import (
+	"context"
+	"encoding/binary"
+
+	"github.com/bougou/go-ipmi/pkg/handlers"
+)
+
+// --- FRU binary helpers ----------------------------------------------------
```
anishbista60:
```suggestion
```

Similar to this.

anishbista60 (dismissed):
Overall LGTM. Thanks for your effort 

/cc @starbops

pkg/ipmi/simulator.go
```diff
@@ -1,262 +1,181 @@
...
+		if err := s.srv.Serve(ctx); err != nil && !errors.Is(err, context.Canceled) {
+			logrus.WithError(err).Error("IPMI server exited with error")
+		}
+	}()
 
-	// Extract username from the 16-byte username field (trim null bytes)
-	username := string(bytes.TrimRight(r.Username[:], "\x00"))
-
-	return s.processSessionChallenge(username)
+	logrus.Infof("IPMI service listening on %s", addr)
```
starbops: This line is duplicated. Either show the message here or https://github.com/kubevirtbmc/kubevirtbmc/blob/f43ac24526688a8125f171a0d7cac3513750428e/pkg/virtbmc/virtbmc.go#L86

starbops (commented):
Works perfectly! Just one nit. Thank you.

pkg/ipmi/simulator.go
halfcrazy: yes i change this in the fru commit, will change later

starbops (approved):

Issue #141, opened 2026-02-21
[TASK] Migrate the Helm chart directory to the standalone chart repository


starbops:
With the introduction of the kubevirtbmc organization, we should move the Helm chart from the main repository to its standalone repository.

FWIW, we need to figure out how to maintain the relationship between the chart generated in #126 and the chart repository.

PR #242, opened 2026-07-06
chore: remove migrated Helm chart

starbops:
Related issue: #141


anishbista60 (approved):
Thanks for the great effort.

Issue #205, opened 2026-06-16
[ENHANCEMENT] Virtual Media E2E Tests with VM-Side Validation


anishbista60:
**Is your feature request related to a problem? Please describe.**

The current virtual media E2E test only verifies that the Redfish client successfully calls `InsertMedia` and `EjectMedia`. It does not validate that the expected changes occur on the VM side.

**Describe the solution you'd like**

Enhance the virtual media E2E test to verify the full workflow:

* A DataVolume is created when `InsertMedia` is called.
* The VM is updated to boot from the attached ISO and restarted.
* The VM successfully boots from the virtual media.
* The DataVolume is deleted when `EjectMedia` is called.
* The corresponding volume entries are removed from the VM spec.

**Describe alternatives you've considered**

Continue validating only the Redfish API responses. However, this does not guarantee that the virtual media workflow is correctly applied to the VM.

**Additional context**

This enhancement would provide true end-to-end validation of the virtual media feature and help catch regressions in VM boot and media cleanup behavior.

PR #244, opened 2026-07-08
feat: enhance the e2e of virtual media mount

anishbista60:
Tracking issue: https://github.com/kubevirtbmc/kubevirtbmc/issues/205, https://github.com/kubevirtbmc/kubevirtbmc/pull/186#issuecomment-4854676456

- This PR made the enhancement for the virtual media mount

### DIFF
--- a/test/virtbmc-agent/agent_helpers.go
+++ b/test/virtbmc-agent/agent_helpers.go
@@ -15,9 +15,10 @@ import (
 	kubescheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/remotecommand"
+	kubevirtv1 "kubevirt.io/api/core/v1"
+	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	kubevirtv1 "kubevirt.io/api/core/v1"
 	bmcv1 "kubevirt.io/kubevirtbmc/api/bmc/v1beta1"
 )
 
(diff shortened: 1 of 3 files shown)


test/virtbmc-agent/agent_test.go
```diff
@@ -461,13 +467,36 @@ var _ = Describe("Agent e2e", Ordered, func() {
 				))
 			})
 
+			It("should create a DataVolume after InsertMedia", func() {
+				verifyDataVolumeExists(ctx, k8sClient, ns, agentVMName)
+			})
```
halfcrazy, line 472: Should we add a check for DV annotation `cdi.kubevirt.io/storage.bind.immediate.requested` here, or maybe PVC bound?
introduced by https://github.com/kubevirtbmc/kubevirtbmc/pull/186/changes

anishbista60: Data volume is created only when PVC get bound.  Check for PVC will just add another step.

Regarding annotation, since it already added at the  actual code in the data volume.  Reverifying it again in the data volume  seems not required. 

We just need to check the functionality that occur after that annotation which is already added.

test/virtbmc-agent/agent_helpers.go
```diff
@@ -402,3 +403,58 @@ func runIPMIInCluster(ctx context.Context, cfg *rest.Config, namespace string, r
...
 	})
 }
+
+func verifyDataVolumeExists(ctx context.Context, k8sClient client.Client, namespace, name string) {
+	Eventually(func() error {
+		dv := &cdiv1.DataVolume{}
+		return k8sClient.Get(ctx, client.ObjectKey{Namespace: namespace, Name: name}, dv)
+	}, agentTestTimeout, agentTestInterval).Should(Succeed(),
+		"DataVolume %s/%s should exist", namespace, name)
+}
```
anishbista60, line 413: I think I need to add the another check for datavolume when it reach progressing state (when it acutally start downloading the ISO image)

test/virtbmc-agent/agent_test.go
halfcrazy: From what I recall, PVCs are created by DV resources. For DV import workflows: if that annotation is omitted, disk provisioning only triggers when the virtual machine powers on and mounts the volume. Once the annotation is added, provisioning happens immediately after the DV is created.

anishbista60: Lol. You are correct, So many things at at same time.I was not able to think properly while writing above comment lol

anishbista60: Data volume start importing the ISO image after PVC get bound.

halfcrazy (approved):

Issue #247, opened 2026-07-17
[ENHANCEMENT] Improve local dev experience for make run against a local cluster


anishbista60:
**Is your enhancement request of an existing feature related to a problem? Please describe.**

When running `make run` against a local (non-in-cluster) Kubernetes cluster fails out of the box. Specifically:

  - The manager crashes on startup because controller-runtime's webhook server requires TLS certs at `/tmp/k8s-webhook-server/serving-certs`, which don't exist outside a real cluster.
  - `make run` doesn't depend on `make install`, so the controller can start against a cluster that doesn't have the required CRDs installed yet.
  - The virtbmc agent pods CrashLoopBackOff with RBAC "forbidden" errors, because the ClusterRole they need is only applied as part of `make deploy`'s full `config/default` bundle — not by `make run`.

  **Describe the solution you'd like**

  - Add an `ENABLE_WEBHOOKS` env var (default `true`) so the webhook server/registration can be explicitly disabled for local runs.
  - Make the `run` target depend on `install` so CRDs are always applied before the controller starts.
  - Add an `install-virtbmc-rbac` make target, wired into `run`, that applies the ClusterRole the agent pods need independently of the full `make deploy` bundle.

starbops:
Thank you for proposing the enhancement. I have some questions that need clarification:

- I believe kubebuilder already respects an environment variable `ENABLE_WEBHOOKS`, which we can utilize directly
- If we wire the `install` target to the `run` target, resources installed, such as the CRD, could remain in the cluster when `make run` is interrupted by `ctrl-c`. This is an undesirable behavior, and developers need to manually clean them up.
- The `install-virtbmc-rbac` target shares the same reason as the previous one

In addition to the points mentioned earlier, we need improved developer documentation to enhance the developer experience.

anishbista60:
> * I believe kubebuilder already respects an environment variable `ENABLE_WEBHOOKS`, which we can utilize directly

ok. Mind sharing where we can find this information ? I don't see these ENV variable in our code base. 

> 
> In addition to the points mentioned earlier, we need improved developer documentation to enhance the developer experience.

haha lol yes

starbops:
Sorry I put it wrong. It's not included in the kubebuilder codebase, it's in their sample code. In the last paragraph from the kubebuilder documentation page https://book.kubebuilder.io/multiversion-tutorial/webhooks.html. And I have a vague impression that kubebuilder scaffold did this in our codebase.

[...]

PR #248, opened 2026-07-17
fix(local-dev): unblock `make run` against a local cluster

anishbista60:
Tracking issue: #247


anishbista60:
Once local controller deploy the bmc pod, there will be image pull back off error. I am thinking to add the image build part and loading to the kind cluster as the part of the make target.

PR #249, opened 2026-07-20
feat(controller): add configurable resource requests for agent pod

### LINKED ISSUE #246: [ENHANCEMENT] VirtBMC Agent pod lack request resource

Some Kubernetes distributions refuse to schedule Pods that lack configured resource requests. 
And pods without resource requests are excluded from this reservation calculation and will be evicted first once node resources are exhausted.

Add a minimal pod resource request. Eg 50m. Or make it a configuration for the controller.

**Additional context**

Add any other context or screenshots about the enhancement request here.

halfcrazy:
Some Kubernetes distributions refuse to schedule Pods without resource requests, and such Pods are excluded from reservation calculation and evicted first under node pressure.

Set default resource requests (cpu: 10m, memory: 128Mi) on the virtBMC agent container, overridable via new controller flags --agent-cpu-request and --agent-memory-request, following the same pattern as --agent-image-name/--agent-image-tag. Values are validated at controller startup.

The default value takes from our environment the real usage.
```
vm-z0bhof2g2zkpibx4yh6f-virtbmc               1/1     Running                  0          5h30m
vm-zgg670jcq8kot1z9qzdi-virtbmc               1/1     Running                  0          5h30m
vm-zujnib1ohefbs6b9qiig-virtbmc               1/1     Running                  0          5h31m
[root@node-3 ~]# kubectl top po vm-zujnib1ohefbs6b9qiig-virtbmc
NAME                              CPU(cores)   MEMORY(bytes)
vm-zujnib1ohefbs6b9qiig-virtbmc   2m           93Mi
[root@node-3 ~]# kubectl top po vm-zgg670jcq8kot1z9qzdi-virtbmc
NAME                              CPU(cores)   MEMORY(bytes)
vm-zgg670jcq8kot1z9qzdi-virtbmc   0m           87Mi
[root@node-3 ~]# kubectl top po vm-z0bhof2g2zkpibx4yh6f-virtbmc
NAME                              CPU(cores)   MEMORY(bytes)
vm-z0bhof2g2zkpibx4yh6f-virtbmc   1m           90Mi
```

Closes #246

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -26,6 +26,7 @@ import (
 	// to ensure that exec-entrypoint and run can make use of them.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
@@ -70,6 +71,8 @@ func main() {
 		tlsOpts              []func(*tls.Config)
 		agentImageName       string
 		agentImageTag        string
+		agentCPURequest      string
+		agentMemoryRequest   string
 	)
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8443", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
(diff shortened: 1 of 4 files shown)


anishbista60 (commented):
Thanks for working on it. 

Would be good way to pass request and limit via controller flag ? 

I would say we could pass these things via BMC Custom resource because of following reason: 

1) Single point of control via BMC CR
2) No need to update the controller manager flag while upgrading the KubeVirtBMC and while adjusting the request and limt
3) No much hassle. 

Happy to see your and @starbops perspective like why with controller flag. 

Thanks !

halfcrazy:
> Thanks for working on it. 
> 
> 
> 
> Would be good way to pass request and limit via controller flag ? 
> 
> 
> 
> I would say we could pass these things via BMC Custom resource because of following reason: 
> 
> 
> 
> 1) Single point of control via BMC CR
> 
> 2) No need to update the controller manager flag while upgrading the KubeVirtBMC and while adjusting the request and limt
> 
> 3) No much hassle. 
> 
> 
> 
> Happy to see your and @starbops perspective like why with controller flag. 
> 
> 
> 
> Thanks ! 
> 
>  
> 
> 
> 
> 
> 
> 

I haven’t added the limit parameter in this iteration. We can expose it as a configuration option if needed (no default value; the setting will be omitted if unspecified).

Regarding your suggestion to expose this parameter via CR configurations, I don’t think it’s necessary. Based on my practical usage experience, this parameter is most likely a global setting and there is no requirement for per-instance customization. Therefore, controlling it through flags will suffice.

anishbista60:
> this parameter is most likely a global setting and there is no requirement for per-instance customization. Therefore, controlling it through flags will suffice.


Sounds good from global perspective.

anishbista60 (approved):
LGTM. Thanks for the PR. 

/cc @starbops for final call. Seem's he is occupied these days. 

Thanks !

starbops:
Thank you for the PR.

I agree that both points are valid. However, at the current stage, we don’t observe any significant variation in resource usage among agents. Agents primarily serve as translators between IPMI/Redfish and Kunernetes API, and they don’t perform complex actions. Therefore, having a global configuration for resource requirements is sufficient for now. If the need arises in the future, we can discuss per-VM configuration through the CRD.

anishbista60 (commented):
Thanks @starbops for your opinion and @halfcrazy for this enhancement. 

Merging it.

PR #254, opened 2026-07-27
security: set restricted-compliant securityContext on virtbmc Pod

anishbista60:
Same as title 

kubevirtbmc/security#2

### DIFF
--- a/internal/controller/virtualmachinebmc/controller.go
+++ b/internal/controller/virtualmachinebmc/controller.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
 	kubevirtv1 "kubevirt.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


starbops (commented):
Thanks for the PR. I don't think the `/tmp` `emptyDir` volume is necessary. Is there a specific reason for adding it?

anishbista60:
> Thanks for the PR. I don't think the `/tmp` `emptyDir` volume is necessary. Is there a specific reason for adding it?

Assuming `readOnlyRootFilesystem` would break the multipart upload helpers in the generated redfish server code initially but didn't realize those helpers aren't actually wired into the real handler


Removing it.

starbops:
> > Thanks for the PR. I don't think the `/tmp` `emptyDir` volume is necessary. Is there a specific reason for adding it?
> 
> Assuming `readOnlyRootFilesystem` would break the multipart upload helpers in the generated redfish server code initially but didn't realize those helpers aren't actually wired into the real handler
> 
> Removing it.

We can surely add it back later once those handlers are implemented and wired in. Thank you.

starbops (approved):
LGTM, thank you! Please feel free to merge it.

PR #256, opened 2026-07-29
fix(virtbmc): handle SIGTERM for graceful agent shutdown

halfcrazy:
Related issue: https://github.com/kubevirtbmc/kubevirtbmc/issues/257

The agent only registered SIGINT via signal.Notify, so when Kubernetes deleted the pod, SIGTERM fell through to the default disposition and killed the process (exit code 143), leaving the pod in Error state. Register SIGTERM as well so the context is cancelled and the agent shuts down gracefully with exit code 0.

Also switch the e2e ipmitool image to kubevirtbmc/ipmitool.

### DIFF
--- a/cmd/virtbmc/main.go
+++ b/cmd/virtbmc/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"syscall"
 
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v2"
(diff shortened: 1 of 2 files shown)


anishbista60 (approved):
LGTM. Thanks for fixing this.

starbops:
We'd better have a GitHub issue tracking the change. This is good for project management.

anishbista60:
Since, it was small fix that's why I ignore this. Yes, you are correct we should follow best practices. 

I have created the issue and closed it. Thanks

Issue #162, opened 2026-03-29
[ENHANCEMENT] Replace Pod with Deployment


anishbista60:
Is your enhancement request of an existing feature related to a problem? Please describe.
Using Pods directly limits scalability, self-healing, and rolling updates.

Describe the solution you'd like
Replace Pod with a Deployment to manage lifecycle, scaling, and updates.

Additional context
Improves reliability and supports rolling updates.

PR #263, opened 2026-07-30
fix(controller): recreate agent pod on template drift

### LINKED ISSUE #262: [BUG] Existing agent instances will not be automatically updated upon version upgrade.
**Describe the bug**

Existing agent instances will not be automatically updated upon version upgrade.

For example
- update agent image via controller args
- add CPU and memory requests
- add probe configuration

**To Reproduce**

**Expected behavior**

Agent pod update if pod template changes.

**Screenshots**

- OS: [e.g. SLES 15 SP5]
- Chart version: [e.g. 0.2.3]
- App version: [e.g. v0.5.3]

**Additional context**

halfcrazy:
Recreate existing virtbmc pods when the agent image name or tag, CPU or memory requests, or probe configuration changes. Stamp the desired pod template hash, so upgrades replace stale agents while preserving Service identity.

As a tradeoff, existing instances without the podtemplatehash will also be recreated once after the version upgrade. Nevertheless, controller updates are most likely accompanied by agent version updates.

fix #262

### DIFF
--- a/internal/controller/virtualmachinebmc/controller.go
+++ b/internal/controller/virtualmachinebmc/controller.go
@@ -23,6 +23,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -454,18 +455,6 @@ func specIPMIEnabled(spec *bmcv1.VirtualMachineBMCSpec) bool {
 	return spec.IPMI != nil && spec.IPMI.Enabled != nil && *spec.IPMI.Enabled
 }
 
-func podIPMIEnabled(pod *corev1.Pod) bool {
-	val, ok := pod.Annotations[EnableIPMIAnnotation]
-	if !ok {
-		return false
-	}
-	enabled, err := strconv.ParseBool(val)
-	if err != nil {
-		return false
-	}
-	return enabled
-}
-
 func (r *VirtualMachineBMCReconciler) patchVirtBMCServicePorts(
 	ctx context.Context,
 	virtualMachineBMC *bmcv1.VirtualMachineBMC,
@@ -486,6 +475,10 @@ func (r *VirtualMachineBMCReconciler) patchVirtBMCServicePorts(
 
 	desired := r.constructServiceFromVirtualMachineBMC(virtualMachineBMC)
 
+	if apiequality.Semantic.DeepEqual(existing.Spec.Ports, desired.Spec.Ports) {
+		return nil
+	}
+
 	patch := client.MergeFrom(existing.DeepCopy())
 	existing.Spec.Ports = desired.Spec.Ports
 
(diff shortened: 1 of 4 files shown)


internal/controller/virtualmachinebmc/controller.go
```diff
@@ -486,6 +475,10 @@ func (r *VirtualMachineBMCReconciler) patchVirtBMCServicePorts(
 
 	desired := r.constructServiceFromVirtualMachineBMC(virtualMachineBMC)
 
+	if apiequality.Semantic.DeepEqual(existing.Spec.Ports, desired.Spec.Ports) {
+		return nil
+	}
+
```
anishbista60, line 481: I'm not sure whether this would be a good way. I also did intially same but later on I switch based on the annotation comparsion. 

see the comment by co-pilot which is valid I  feel : https://github.com/kubevirtbmc/kubevirtbmc/pull/181#discussion_r3254408266

internal/controller/virtualmachinebmc/pod_template.go
```diff
@@ -0,0 +1,40 @@
...
+		return err
+	}
+
+	sum := sha256.Sum256(data)
+	if pod.Annotations == nil {
+		pod.Annotations = map[string]string{}
+	}
+	pod.Annotations[PodTemplateHashAnnotation] = hex.EncodeToString(sum[:])
+	return nil
+}
```
anishbista60, line 40: If I change the image name manually then does  the hash will change ?

internal/controller/virtualmachinebmc/controller.go
```diff
@@ -593,7 +583,15 @@ func (r *VirtualMachineBMCReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 	}
 
-	deleted, err := r.reconcileIPMIChange(ctx, &virtualMachineBMC)
+	pod := r.constructPodFromVirtualMachineBMC(&virtualMachineBMC)
+	if err := ctrl.SetControllerReference(&virtualMachineBMC, pod, r.Scheme); err != nil {
+		return ctrl.Result{}, err
+	}
+	if err := stampPodTemplateHash(pod); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to hash virtBMC Pod template: %w", err)
+	}
```
anishbista60, line 592: Since, we will be replacing pod with deployment in future these stamp part will not be necessary cause the pod lifecycle will be managed by rs and rs will be managed by deployement

internal/controller/virtualmachinebmc/pod_template.go
halfcrazy: ~~Image field in `pod.Spec`, so yes.~~

No, the hash was calculated on creation.

internal/controller/virtualmachinebmc/controller.go
halfcrazy: Alright, next we need to handle the scenario where the Deployment gets updated automatically.

anishbista60 (commented):
Thanks for PR and working on it. I feel this issue will be automatically solve while working on https://github.com/kubevirtbmc/kubevirtbmc/issues/162

Till then we can hold it ? WDYT ? 

/cc @halfcrazy @starbops

internal/controller/virtualmachinebmc/controller.go
anishbista60: Wait my question is wrong. 

Why you are adding this extra check ? Will this be required under any scenario ? Mind sharing that ?

halfcrazy: Because this function gets triggered on every reconcile, I explicitly avoid sending empty patches.

anishbista60: Thank you. Sounds good .

halfcrazy:
OK, let's focus on #158

Issue #265, opened 2026-07-31
[BUG] E2E tests should pin a known-good KubeVirt version instead of tracking stable.txt


anishbista60:
## Summary

`InstallKubeVirt()` in [`test/util/helper.go`](../blob/main/test/util/helper.go) always installs whatever KubeVirt version `stable.txt` currently points to:

```go
kubeVirtStableVersion = "https://storage.googleapis.com/kubevirt-prow/release/kubevirt/kubevirt/stable.txt"
```

KubeVirt `v1.9.0` was released the night of 2026-07-30, and `stable.txt` now resolves to it. As a result, `make local-e2e-test` is currently **failing** , unrelated to any change in this repo.

Meanwhile, this project's KubeVirt API is pinned to `v1.8.4` (`go.mod`, `Makefile:42` `KUBEVIRT_API_VERSION`), so the e2e install target and the API we build against have drifted apart on every KubeVirt release.

## Root cause of the current failure

KubeVirt `v1.9.0` shipped [PR #17405](https://github.com/kubevirt/kubevirt/pull/17405), which **enables all Beta feature gates by default**. `ImageVolume` is a Beta gate, so it is now on by default with no opt-in.

When `ImageVolume` is active, `virt-controller` builds the containerDisk init container to exec `/container-disk-binary/usr/bin/container-disk` via the new ImageVolume code path. upstream fix (see [kubevirt#17460](https://github.com/kubevirt/kubevirt/issues/17460), which hit the identical symptom). Our e2e kind cluster is pinned to `KIND_K8S_VERSION = v1.34.8` (`Makefile:22`), one minor version short.

Effect: every containerDisk-backed VMI's `volumecontainerdisk` init container fails immediately:

```
Error: failed to create containerd task: failed to create shim task: OCI runtime create failed:
runc create failed: unable to start container process: error during container init:
exec: "/container-disk-binary/usr/bin/container-disk": stat /container-disk-binary/usr/bin/container-disk: no such file or directory
```

Since the e2e `testvm` VM's `RunStrategy` is `Always`, the VM controller deletes and recreates the VMI/pod in a continuous loop (new pod every ~15-20s) rather than a simple per-pod CrashLoopBackOff, and `make local-e2e-test` times out waiting for the VMI to reach `Running`.

## Proposal

[...]

PR #266, opened 2026-07-31
fix(e2e): pin KubeVirt install version instead of tracking stable.txt

anishbista60:
Tracking issue: #265

### DIFF
--- a/test/util/helper.go
+++ b/test/util/helper.go
@@ -14,7 +14,7 @@ import (
 
 const (
 	certManagerURLFmt      = "https://github.com/jetstack/cert-manager/releases/download/%s/cert-manager.yaml"
-	kubeVirtVersion        = "v1.8.3"
+	kubeVirtVersion        = "v1.8.4"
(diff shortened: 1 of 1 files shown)


test/util/helper.go
```diff
@@ -14,7 +14,7 @@ import (
 
 const (
 	certManagerURLFmt      = "https://github.com/jetstack/cert-manager/releases/download/%s/cert-manager.yaml"
-	kubeVirtStableVersion  = "https://storage.googleapis.com/kubevirt-prow/release/kubevirt/kubevirt/stable.txt"
+	kubeVirtVersion        = "v1.8.3"
```
starbops: Any reason that we don't use v1.8.4?

anishbista60: sorry it should be 1.8.4

pohanhuang (approved):
LGTM

starbops (approved):
LGTM, thanks for the fix!

Issue #264, opened 2026-07-31
[ENHANCEMENT] Profile and reduce the virtbmc agent memory footprint


starbops:
**Is your enhancement request of an existing feature related to a problem? Please describe.**

Each virtbmc agent pod currently reports a working set of 87 to 93 MiB. Because KubeVirtBMC maps agent pods to VMs one-to-one, this cost is multiplied by the number of managed VMs. On a 500-VM cluster the agents alone would account for roughly 45 GiB.

The cost is not only runtime. PR #249 derived the default `--agent-memory-request` of 128Mi directly from these observed numbers, so the footprint is now what the scheduler reserves per VM. Until we know why an agent needs 90 MiB, we cannot justify lowering that request, and VM density per node stays capped by it.

The agent is a translator between IPMI/Redfish and the Kubernetes API. It does not perform heavy work, and a Go HTTP server baseline is typically 10 to 20 MiB. 90 MiB is high for what the agent does, but we currently have no profile that explains the gap. That is the gap this issue addresses.

**Describe the solution you'd like**

The scope of this issue is diagnosis, not a fix. The deliverable is a profile that shows where the memory goes, plus a repeatable measurement procedure. Implementation work gets its own issues once the profile points somewhere.

Steps:

1. Split the agent's RSS into anonymous and file-backed pages by reading `/proc/<pid>/smaps_rollup` inside the agent container. This determines whether the cost is Go heap and stacks or mapped binary text.
2. Capture `runtime.MemStats` (HeapAlloc, HeapSys, Sys) and a pprof heap profile from a running agent, both at idle and under IPMI/Redfish load. `kubectl top` reports `container_memory_working_set_bytes`, which is not the Go heap, so the two need to be measured separately.
3. Determine whether the footprint scales with cluster size by comparing one agent on a cluster with a handful of VMs against one agent on a cluster with many. If the number grows, the cause is client-side caching and it dominates every other consideration.
4. Record the agent binary size, and the effective `GOMAXPROCS` and `GOMEMLIMIT` as observed by the process inside the container.

Hypotheses to confirm or rule out, in rough order of expected impact:

- Kubernetes client caches. If the agent runs an informer or a cached client that is not scoped by field selector to its own VM, every agent caches every relevant object in the cluster. This is the only candidate that explains growth with cluster size.
- Unbounded Go heap. With no `GOMEMLIMIT` set and the default `GOGC=100`, the heap grows to roughly twice the live set and pages are returned to the OS lazily.
[...]

PR #269, opened 2026-08-03
perf(agent): strip unimplemented Redfish routes to cut agent memory

halfcrazy:
The generated Redfish server registers one gorilla/mux route per spec operation (4034 routes), and mux compiles every pattern into a regexp that stays live for the process lifetime. Only 17 operations have real implementations; the rest answer 501. The compiled regexps accounted for ~50MB of live heap per agent pod, the dominant share of the 87-93Mi working set reported in #264.

Register only routes backed by a real implementation:

- routeFilter drops stub-only routes from the generated route table; unimplemented endpoints now return 404 (what real BMCs answer for unsupported services) instead of 501.
- pkg/redfish/strip-redfish-routes derives the implemented set from api_service.go via go/ast (methods not returning the generated 501 stub) and runs through go:generate, so the set stays in sync as endpoints get implemented. Wired into make generate / docker-build / docker-buildx so local builds, image builds and the PR freshness check all cover it.

Measured against a 433-CRD / 119-VM cluster: agent RSS 126MB -> 46.5MB, live heap 66.6MB -> 6.8MB; in-cluster working set is expected at ~30-40Mi (file-backed pages get reclaimed under memory pressure). Lower DefaultAgentMemoryRequest 128Mi -> 50Mi; the agent pod sets no memory limit, so the request only affects scheduling and eviction ranking, and steady-state usage stays below it. On a 500-VM cluster this cuts the memory reserved for agents from 64Gi to 25Gi.


anishbista60:
I will take a look during weekend and get it approve.

pkg/redfish/strip-redfish-routes/main.go
```diff
@@ -0,0 +1,101 @@
...
+// must not, otherwise it would be dropped from the route table.
+func isNotImplementedStub(fn *ast.FuncDecl) bool {
+	stub := false
+	ast.Inspect(fn.Body, func(n ast.Node) bool {
+		if stub {
+			return false
+		}
+		if sel, ok := n.(*ast.SelectorExpr); ok && sel.Sel.Name == "StatusNotImplemented" {
+			stub = true
+		}
```
anishbista60: Can we add on more layer of check so that only `http.StatusNotImplemented` matches ? 

Currently, it check `*.StatusNotImplemented`

Makefile
```diff
@@ -164,7 +168,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 # (i.e. docker build --platform linux/arm64). However, you must enable docker buildKit for it.
 # More info: https://docs.docker.com/develop/develop-images/build_enhancements/
 .PHONY: docker-build
-docker-build: ## Build docker images with the manager and the agent respectively.
+docker-build: generate-implemented-routes ## Build docker images with the manager and the agent respectively.
```
anishbista60, line 171: Nit: 

It means image binary only includes the routes which are only in use ?

pkg/redfish/implemented_routes_gen.go
```diff
@@ -0,0 +1,27 @@
...
+	"RedfishV1SessionServiceSessionsPost":                                                    true,
+	"RedfishV1SessionServiceSessionsSessionIdDelete":                                         true,
+	"RedfishV1SessionServiceSessionsSessionIdGet":                                            true,
+	"RedfishV1SystemsComputerSystemIdActionsComputerSystemResetPost":                         true,
+	"RedfishV1SystemsComputerSystemIdActionsComputerSystemSetDefaultBootOrderPost":           true,
+	"RedfishV1SystemsComputerSystemIdGet":                                                    true,
+	"RedfishV1SystemsComputerSystemIdOperatingSystemGet":                                     true,
+	"RedfishV1SystemsComputerSystemIdPatch":                                                  true,
+	"RedfishV1SystemsGet":                                                                    true,
+}
```
anishbista60, line 27: What if we would add the more routes in future. I believe we need to run `make generate` manually ?

anishbista60 (commented):
Thanks for the PR. I have left few questions/changes.

pkg/redfish/strip-redfish-routes/main.go
halfcrazy: Good point. The scan only sees the generated `api_service.go`, whose stubs always reference `http.StatusNotImplemented` — but a future real implementation could legitimately use it (e.g. returning 501 for an unsupported sub-resource) and would silently lose its route. I will tighten the match to require `sel.X` to be the `http` ident, so only the generated-stub shape counts.

Makefile
halfcrazy: Not exactly — all generated handler methods (including the 501 stubs) still compile into the binary. Stripping only affects which routes get registered with gorilla/mux at runtime, so the saving is process heap (mux no longer compiles ~4000 patterns into regexps, ~50MB), not binary/image size. Removing the unimplemented routes from the binary itself would require trimming the Redfish OpenAPI schema fed to the generator, which is out of scope here. The `generate-implemented-routes` dependency on docker-build just ensures the checked-in `implemented_routes_gen.go` is refreshed before the image is built, so a stale route set cannot be packaged.

pkg/redfish/implemented_routes_gen.go
halfcrazy: Yes — after implementing a new route you run `make generate` (or `go generate ./pkg/redfish`) once; it is a generated file, so no manual editing. It is already enforced in CI: the "Verify generated files are up-to-date" step in `pull-request.yaml` runs `make manifests generate generate-kubevirt-crd` and fails with a hint if the tree is dirty, so a forgotten regeneration is caught in the PR check. `docker-build`/`docker-buildx` also regenerate first, so image builds never carry a stale route set.

anishbista60: Can you mention this part to the contributing.md guide ?

halfcrazy: done

anishbista60 (approved):
LGTM. Thanks !

/cc @starbops

pkg/redfish/implemented_routes_test.go
starbops, line 1: Do we need to manually update this whenever there are updates in the generated `implemented_routes_gen.go`?

starbops (commented):
Thank you for the enhancement!

pkg/redfish/implemented_routes_test.go
halfcrazy: No. This is only a cover for one specific type of data, not full routes.

starbops: So it's just sort of a sanity check?

halfcrazy: yes

starbops (approved):
LGTM, thank you!

anishbista60:
@halfcrazy Feel free to merge it. Thanks !

PR #276, opened 2026-08-08
Feat: Replace pod with deployment

anishbista60:
Tracking issue: #162, #262

- Changed the pod with deployment
- Refractor the controller test 
- Refractor the E2E. 
- Move the labels to api level so that both controller and agent can consume. 

> [!NOTE]
> As a part this PR the default value of the bmc deployment replica is set to 1. User are not allow to set the replicas to 2 . We need to introduce the new field in the CRD. I will create the sub task for it in issue 136

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: anish60/virtbmc-controller
-  newTag: deployv1
+  newName: kubevirtbmc/virtbmc-controller
+  newTag: dev
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: kubevirtbmc/virtbmc-controller
-  newTag: dev
+  newName: anish60/virtbmc-controller
+  newTag: deployv1
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -27,6 +27,18 @@ rules:
   - list
   - watch
 - apiGroups:
+  - apps
+  resources:
+  - deployments
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
   - bmc.kubevirt.io
   resources:
   - virtualmachinebmcs
--- a/internal/controller/virtualmachinebmc/bmc_controller.go
+++ b/internal/controller/virtualmachinebmc/bmc_controller.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"strconv"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -30,7 +31,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
-	"k8s.io/utils/ptr"
 	kubevirtv1 "kubevirt.io/api/core/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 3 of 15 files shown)


internal/controller/virtualmachinebmc/deployment.go
```diff
@@ -0,0 +1,297 @@
...
+	deployment := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      deploymentName,
+			Namespace: virtualMachineBMC.Namespace,
+			Labels:    labels,
+		},
+		Spec: appsv1.DeploymentSpec{
+			Replicas: int32Ptr(defaultDesiredReplicas),
+			Selector: &metav1.LabelSelector{
+				MatchLabels: labels,
```
halfcrazy, line 69: Do we allow two BMCs to refer to the same VM? If so, the labels may conflict

internal/controller/virtualmachinebmc/bmc_controller.go
```diff
@@ -500,33 +339,34 @@ func (r *VirtualMachineBMCReconciler) patchVirtBMCServicePorts(
...
 		return false, nil
 	}
 
-	log.Info("enableIPMI changed, replacing pod and patching service",
+	log.Info("enableIPMI changed, replacing deployment and patching service",
 		"currentHasIPMI", currentHasIPMI,
 		"desiredHasIPMI", desiredHasIPMI)
 
-	if err := r.deleteVirtBMCPod(ctx, virtualMachineBMC); err != nil {
+	if err := r.deleteVirtBMCDeployment(ctx, virtualMachineBMC); err != nil {
```
halfcrazy: IPMI change: we don't need to delete deploy; SSA is enough. Just take care of the service

internal/controller/virtualmachinebmc/deployment.go
```diff
@@ -0,0 +1,297 @@
...
+	"k8s.io/utils/ptr"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	bmcv1 "kubevirt.io/kubevirtbmc/api/bmc/v1beta1"
+)
+
+const (
+	defaultDesiredReplicas int32 = 1
```
halfcrazy, line 41: Do we need more than 1 pod to serve Redfish? It's stateful(Redfish session).
During an upgrade, having two replicas is an acceptable intermediate state; however, dual‑replica runtime is not recommended in the long‑term.

halfcrazy (changes requested):

internal/controller/virtualmachinebmc/deployment.go
```diff
@@ -0,0 +1,297 @@
...
+		Spec: appsv1.DeploymentSpec{
+			Replicas: int32Ptr(defaultDesiredReplicas),
+			Selector: &metav1.LabelSelector{
+				MatchLabels: labels,
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: labels,
+					Annotations: map[string]string{
+						EnableIPMIAnnotation: strconv.FormatBool(specIPMIEnabled(&virtualMachineBMC.Spec)),
```
halfcrazy:
```
if secretHash != "" {
    annotations["bmc.kubevirt.io/secret-hash"] = secretHash
}
```
Then we don't need to roll out manually

internal/controller/virtualmachinebmc/deployment.go
```diff
@@ -0,0 +1,297 @@
...
+		}
+		log.Error(err, "unable to delete virtBMC Deployment", "deployment", deployment)
+		return err
+	}
+
+	log.Info("deleted virtBMC Deployment", "deployment", deployment)
+	return nil
+}
+
+func (r *VirtualMachineBMCReconciler) rolloutRestartVirtBMCDeployment(ctx context.Context, virtualMachineBMC *bmcv1.VirtualMachineBMC) error {
```
halfcrazy: https://github.com/kubevirtbmc/kubevirtbmc/pull/276/changes#r3740559319

internal/controller/virtualmachinebmc/deployment.go
```diff
@@ -0,0 +1,297 @@
...
+	if err := r.Delete(ctx, pod); err != nil && !apierrors.IsNotFound(err) {
+		log.Error(err, "unable to delete legacy virtBMC Pod", "pod", podName)
+		return err
+	}
+
+	log.Info("deleted legacy virtBMC Pod", "pod", podName)
+	return nil
+}
+
+func int32Ptr(i int32) *int32 {
```
halfcrazy: k8s.io/utils/ptr
ptr.To

test/util/e2e_resources.go
```diff
@@ -25,14 +26,35 @@ const (
...
+func AgentDeploymentKey(namespace string) types.NamespacedName {
+	return types.NamespacedName{Name: E2EAgentDeploymentName, Namespace: namespace}
+}
+
+func AgentServiceKey(namespace string) types.NamespacedName {
+	return types.NamespacedName{Name: E2EAgentDeploymentName, Namespace: namespace}
+}
+
+var agentPodLabels = client.MatchingLabels{
+	"kubevirt.io/virtualmachinebmc-name": E2EBMCName,
```
halfcrazy: bmcv1.VirtualMachineBMCNameLabel / bmcv1.VMNameLabel

internal/controller/virtualmachinebmc/deployment.go
anishbista60: Let's stick with 1 pod. 2 pod during rolling update for few second should be acceptable.

internal/controller/virtualmachinebmc/deployment.go
anishbista60: As per the design principle, one bmc to one unique VM. Probably, we should reject this scenario  at webhook level

WDYT ? 

/cc @starbops  as well.

internal/controller/virtualmachinebmc/bmc_controller.go
```diff
@@ -499,42 +338,30 @@ func (r *VirtualMachineBMCReconciler) patchVirtBMCServicePorts(
...
-	if err := r.deleteVirtBMCPod(ctx, virtualMachineBMC); err != nil {
-		return false, err
-	}
-
-	if err := r.patchVirtBMCServicePorts(ctx, virtualMachineBMC); err != nil {
-		return false, err
-	}
-
-	return true, nil
+	return r.patchVirtBMCServicePorts(ctx, virtualMachineBMC)
```
halfcrazy: Suggest always patch, dont matter whether IPMI changed

anishbista60: sounds good.

halfcrazy:
thank you. lgtm

halfcrazy (approved):

internal/controller/virtualmachinebmc/deployment.go
starbops: What prevent us to have multiple agent pods for any specific VM currently? I suppose none, since we haven't introduced any feature that requires state management or data persistence in pod. We may have controllers implemented in agents in the future but that can be handled by adding leader election.

Let me know if I miss anything.

starbops: Right, there is. I just saw this https://github.com/kubevirtbmc/kubevirtbmc/pull/276#discussion_r3740550187.

Let's keep it single replica for now.

internal/controller/virtualmachinebmc/deployment.go
```diff
@@ -0,0 +1,300 @@
...
+
+const (
+	defaultDesiredReplicas int32 = 1
+	fieldManager                 = "kubevirtbmc-controller"
+)
+
+func (r *VirtualMachineBMCReconciler) createVirtBMCDeployment(virtualMachineBMC *bmcv1.VirtualMachineBMC, secretHash string) *appsv1.Deployment {
+
+	serviceAccountName := fmt.Sprintf("%s-virtbmc", virtualMachineBMC.Spec.VirtualMachineRef.Name)
+	deploymentName := fmt.Sprintf("%s-virtbmc", virtualMachineBMC.Spec.VirtualMachineRef.Name)
```
halfcrazy, line 48: The deployment name is constructed with the VM name. @starbops 
So 2 bmc for 1 VM result in 1 deploy. If one needs IPMI and the other does not, it may keep it inconsistent.

starbops: Good point. I suppose you're referring to the case where a user creates multiple `VirtualMachineBMC` CRs that all map to the same VM, correct? We currently don't handle this case very well.

I think we should append a suffix to the generated agent Deployment name so that multiple agent Deployments can be created for the same VM. That said, I'm not sure why someone would want to do this in practice. Would you mind elaborating on the use case?

anishbista60: Yes, @starbops all these discussion refers to the case where two BMC to one VM. 

that's the reason I suggested to reject that case at webhook level

https://github.com/kubevirtbmc/kubevirtbmc/pull/276#discussion_r3740544538
https://github.com/kubevirtbmc/kubevirtbmc/pull/276#discussion_r3740550187

halfcrazy: This scenario doesn’t exist in real‑world usage; I spotted it while reading the code.

starbops: Got it. Despite limiting the replicas to 1, do you think we should also enforce that each VM can be referenced by at most one `VirtualMachineBMC`? If so, we don't need to bother adding a suffix to the agent Deployment name. If not, we should account for cases where multiple `VirtualMachineBMC` resources reference the same VM.

anishbista60: > do you think we should also enforce that each VM can be referenced by at most one `VirtualMachineBMC`? 

I would say one to one mapping should be fine .

 > If so, we don't need to bother adding a suffix to the agent Deployment name. If not, we should account for cases where multiple `VirtualMachineBMC` resources reference the same VM.
 
 I think we should reject this at webhook level and if we want granular control then we could add that case at the controller level (If there is already BMC present for any reference VM please skip it)that would be straight forward.
 
 Just my thoughts.

starbops: Same thought here. We should block the creation of additional `VirtualMachineBMC` resources for any VM that is already referenced, ideally at the webhook level.

The controller logic should account for this case as well. As you mentioned, we could either skip reconciling any additional `VirtualMachineBMC` resources and emit a clear log message, or reflect the conflict in their status conditions with a meaningful reason and message.

anishbista60: > ideally at the webhook level.

Yep 

> 
> The controller logic should account for this case as well. As you mentioned, we could either skip reconciling any additional `VirtualMachineBMC` resources and emit a clear log message, or reflect the conflict in their status conditions with a meaningful reason and message.

Ideal way would be skip reconcilation and emit a clear log. Again introducing status field might be overkill because  things will be protected at webhook level then controller will take care. This scenario will exist rarely and for that rare case we already have safe guard at webhook level.

anishbista60: I have created issue to track this part. Thank you everyone for your all suggestion and feeback.

config/manager/kustomization.yaml
starbops, line 1: Could we drop this change?

config/rbac/role.yaml
```diff
@@ -26,6 +26,18 @@ rules:
...
+  resources:
+  - deployments
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
starbops, line 40: With the Deployment rule added, could we drop the Pod rule? Or is that still required?

starbops (commented):
Thank you for the great PR. This is a long-awaited one. It generally looks good to me. Just one suggestion and few nits:

If we want to support a smooth upgrade experience by handling the legacy agent pods removal, it's encouraged that having an E2E test case added to increase the confidence. At least it should be covered with integration tests at the controller level. I understand that this piece of logic is considered temporarily, but it's still good to have that covered, unless we put this as a breaking change.

config/rbac/role.yaml
starbops: Just realized that the controller needs to delete legacy agent pods... so it's still required, at least for now!

anishbista60:
> If we want to support a smooth upgrade experience by handling the legacy agent pods removal, it's encouraged that having an E2E test case added to increase the confidence. At least it should be covered with integration tests at the controller level. I understand that this piece of logic is considered temporarily, but it's still good to have that covered, unless we put this as a breaking change.

Yes, that would be good. But I already tested locally where there are multiple VM running with BMC across different ns. It works perfectly. 

I can still add it not an issue for me but it needs to be cleanup again in future. That's the reason I skipped in test part. 

What's your final call ? 

1) Add it now and remove later 
2) Test in local carefully and move forward on that basis. 


/cc @starbops @halfcrazy

config/manager/kustomization.yaml
anishbista60: How that changes occur ? I'm not sure. It is deletion and addition of same thing.

halfcrazy:
> > If we want to support a smooth upgrade experience by handling the legacy agent pods removal, it's encouraged that having an E2E test case added to increase the confidence. At least it should be covered with integration tests at the controller level. I understand that this piece of logic is considered temporarily, but it's still good to have that covered, unless we put this as a breaking change.
> 
> Yes, that would be good. But I already tested locally where there are multiple VM running with BMC across different ns. It works perfectly.
> 
> I can still add it not an issue for me but it needs to be cleanup again in future. That's the reason I skipped in test part.
> 
> What's your final call ?
> 
> 1. Add it now and remove later
> 2. Test in local carefully and move forward on that basis.
> 
> /cc @starbops @halfcrazy

I think manual testing is sufficient. For the upgrade‑scenario E2E tests, we also manually construct old‑version Pods (unless we actually deploy a real legacy release, which would be overly complex).

starbops (approved):
Make sense. Let's merge this. Thank you!

anishbista60:
Thank you all your valuable review/suggestion ❤️

Issue #284, opened 2026-08-24
[ENHANCEMENT] Automate CRD sync to kubevirtbmc/chart


anishbista60:
**Is your enhancement request of an existing feature related to a problem? Please describe.**

Whenever a CRD field changes in `config/crd/bases/bmc.kubevirt.io_virtualmachinebmcs.yaml` in this repo, someone has to manually copy the updated file over to the [kubevirtbmc/chart](https://github.com/kubevirtbmc/chart) repo (at `charts/kubevirtbmc/crds/`) and open a PR there by hand. This manual step is easy to forget, and the chart's vendored CRD has already drifted out of sync with the operator's CRD as a result.

**Describe the solution you'd like**

A CI workflow in this repo that detects changes to `config/crd/bases/**` on merge to `main`, copies the updated CRD YAML(s) into `kubevirtbmc/chart` at `charts/kubevirtbmc/crds/`, and opens a PR against that repo with the change — so a human still reviews/merges it, but no longer has to do the copy-paste manually.

**Describe alternatives you've considered**

- Keep doing it manually with a checklist/reminder — doesn't scale and is what's causing the drift today.
- Auto-merge the generated PR — rejected, since chart changes should still get human review before merging.

PR #286, opened 2026-08-25
chore: add ci that sync crd betn main repo and chart repo

anishbista60:
- Tracking issue: #284

### DIFF
--- a/.github/workflows/sync-crds-to-chart.yml
+++ b/.github/workflows/sync-crds-to-chart.yml
@@ -1,0 +1,58 @@
+name: Sync CRDs to chart repo
+
+on:
+  push:
+    branches:
+    - main
+    paths:
+    - config/crd/bases/**
+
+permissions:
+  contents: read
+
+jobs:
+  sync-crds:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out kubevirtbmc
+      uses: actions/checkout@v4
+      with:
+        path: kubevirtbmc
+
+    - name: Check out chart
+      uses: actions/checkout@v4
+      with:
+        repository: kubevirtbmc/chart
+        token: ${{ secrets.CHART_REPO_TOKEN }}
+        path: chart
+
+    - name: Copy CRDs into chart repo
+      run: |
+        cp kubevirtbmc/config/crd/bases/*.yaml chart/charts/kubevirtbmc/crds/
+
+    - name: Open PR against chart repo
+      env:
+        GH_TOKEN: ${{ secrets.CHART_REPO_TOKEN }}
+        COMMIT_SHA: ${{ github.sha }}
+        COMMIT_MESSAGE: ${{ github.event.head_commit.message }}
+      run: |
+        cd chart
+        if git diff --quiet; then
+          echo "No CRD changes to sync"
+          exit 0
+        fi
+
+        BRANCH="sync-crds-${COMMIT_SHA}"
+        git config user.name "kubevirtbmc-bot"
+        git config user.email "actions@github.com"
+        git checkout -b "$BRANCH"
+        git add charts/kubevirtbmc/crds/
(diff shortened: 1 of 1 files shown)


halfcrazy:
Is it feasible to package with RBAC included? This does not involve any variables.

anishbista60:
Yes, that needs to be addressed as well but please check the discord discussion.

halfcrazy (dismissed):

.github/workflows/sync-crds-to-chart.yml
```diff
@@ -0,0 +1,58 @@
...
+      run: |
+        cd chart
+        if git diff --quiet; then
+          echo "No CRD changes to sync"
+          exit 0
+        fi
+
+        BRANCH="sync-crds-${COMMIT_SHA}"
+        git config user.name "kubevirtbmc-bot"
+        git config user.email "actions@github.com"
```
starbops: Let use the following identify for the bot:

```
BOT_NAME = "KubeVirtBMC Bot"
BOT_EMAIL = "321044191+kubevirtbmc-bot@users.noreply.github.com"
```

.github/workflows/sync-crds-to-chart.yml
```diff
@@ -0,0 +1,58 @@
...
+        cp kubevirtbmc/config/crd/bases/*.yaml chart/charts/kubevirtbmc/crds/
+
+    - name: Open PR against chart repo
+      env:
+        GH_TOKEN: ${{ secrets.CHART_REPO_TOKEN }}
+        COMMIT_SHA: ${{ github.sha }}
+        COMMIT_MESSAGE: ${{ github.event.head_commit.message }}
+      run: |
+        cd chart
+        if git diff --quiet; then
```
starbops: If a new file is introduced, such as a new CRD, `git diff --quiet` won't catch it.

.github/workflows/sync-crds-to-chart.yml
```diff
@@ -0,0 +1,58 @@
...
+        token: ${{ secrets.CHART_REPO_TOKEN }}
+        path: chart
+
+    - name: Copy CRDs into chart repo
+      run: |
+        cp kubevirtbmc/config/crd/bases/*.yaml chart/charts/kubevirtbmc/crds/
+
+    - name: Open PR against chart repo
+      env:
+        GH_TOKEN: ${{ secrets.CHART_REPO_TOKEN }}
```
starbops, line 35: Just curious, is `GH_TOKEN` only used by the `gh` command?

starbops (commented):
Thank you for initiating the automation between the main and chart repos. I've already set up the secret for the workflow to run.

A few nits down below.

.github/workflows/sync-crds-to-chart.yml
anishbista60: yes

starbops: Then I suppose the `git push` command would fail. Please correct me if I'm wrong.

anishbista60: It won't fail

Now in this case, `actions/checkout` persists credentials by default `(persist-credentials: true)`, so the token passed in the "Check out chart" step is already wired into `chart/.git/config` and used by any later git command there. `GH_TOKEN` is separate, it's only read by the gh pr create call.

anishbista60: docs link: https://github.com/actions/checkout/blob/f548e57e544e1ff5a4c46bf1e1b8685f8e4a348a/README.md#checkout-v4

Check the line  "The auth token is persisted in the local git config"

starbops: Awesome! Thanks for clarifying.

starbops (approved):
LGTM, thank you!

anishbista60:
@halfcrazy Are you here ? or else I will merge it. 

Edited: I'm merging it. Never mind  haha :)

Issue #275, opened 2026-08-07
[BUG] Redfish Advertise unimplemented links in server root


halfcrazy:
**Describe the bug**

When working on #269 using ast to generate non-stub routes (make unimplemented route response 404 instead of 501; both 501 and 404 are incorrect), I took a look at https://github.com/kubevirtbmc/kubevirtbmc/issues/70#issuecomment-2692337543, and I think there is a bug/improvement with server root advertise; it advertises the unimplemented links as well.

https://github.com/kubevirtbmc/kubevirtbmc/blob/a5ff6712fc3799e7cc3df641821ef4e780008503/pkg/redfish/handler.go#L69-L101

**To Reproduce**

**Expected behavior**

Advertise only implemented links in the service root

**Screenshots**

**Environment (please complete the following information):**

- OS: [e.g. SLES 15 SP5]
- Kubernetes distribution/version: [e.g. RKE2 v1.31.7+rke2r1]
- Chart version: [e.g. 0.2.3]
- App version: [e.g. v0.5.3]

**Additional context**

I'd suggest using ast to generate the server root in the same way as #269

starbops:
We need to be careful with this because, IIRC, some provisioning software may rely on specific links being present.

Issue #70, opened 2025-02-07
[BUG] Redfish: Attribute Links/ManagedBy is missing from the resource


lentzi90:
**Describe the bug**

I am trying to use kubevirtbmc with [Bare Metal Operator](https://github.com/metal3-io/baremetal-operator) and Ironic.
So far I have managed to get the connection and authentication to work. I.e. Ironic is able to reach the Redfish API.
However, it then gets stuck because it expects to find Links/ManagedBy attribute for the System. Here is how it looks in the logs:

```
2025-02-07 07:50:44.639 1 DEBUG sushy.connector [None req-9282022a-d73b-4b77-a90f-f8ac968b8909 - - - - - -] HTTP request: GET http://default-testvm-virtbmc.kubevirtbmc-system.svc/redfish/v1/Systems/1; headers: {'OData-Version': '4.0', 'Accept-Encoding': 'identity'}; body: None; blocking: False; timeout: 60; session arguments: {}; _op /usr/lib/python3.9/site-packages/sushy/connector.py:157
2025-02-07 07:50:44.642 1 DEBUG sushy.connector [None req-9282022a-d73b-4b77-a90f-f8ac968b8909 - - - - - -] HTTP response for GET http://default-testvm-virtbmc.kubevirtbmc-system.svc/redfish/v1/Systems/1: status code: 200 _op /usr/lib/python3.9/site-packages/sushy/connector.py:306
[...]

PR #288, opened 2026-08-25
ci(redfish): run Redfish-Interop-Validator in CI against fake clients

halfcrazy:
Serve the real emulator on in-memory fake clients (hack/redfish/interopserver), validate PR head vs base ref, and post a markdown diff of introduced/fixed errors to the step summary; full HTML reports as artifacts. Ironic's OpenStackIronicProfile v1.2.0 is vendored for offline runs.

This PR is a prerequisite for #275 
ref https://github.com/kubevirtbmc/kubevirtbmc/issues/70#issuecomment-2692337543

### DIFF
--- a/.github/workflows/redfish-interop.yml
+++ b/.github/workflows/redfish-interop.yml
@@ -1,0 +1,109 @@
+name: Redfish Interop
+
+# Runs the DMTF Redfish-Interop-Validator against the Redfish API served by
+# hack/redfish/interopserver (the real VirtualMachineResourceManager on
+# in-memory fake clients — no cluster, no KubeVirt). On PRs the base ref is
+# validated too, and the step summary diffs the two runs: regressions
+# introduced vs. failures fixed by the PR.
+# Gating: the job fails when the PR introduces NEW failures vs. the base run
+# (--fail-on-introduced). Without a base run (workflow_dispatch, or a base ref
+# predating the interop tooling) the report is informational only.
+# Local equivalent: make redfish-interop.
+
+on:
+  pull_request:
+    paths:
+      - "pkg/redfish/**"
+      - "pkg/resourcemanager/**"
+      - "pkg/generated/redfish/**"
+      - "pkg/session/**"
+      - "hack/redfish/**"
+      - ".github/workflows/redfish-interop.yml"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: redfish-interop-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  PROFILE: hack/redfish/profiles/OpenStackIronicProfile.v1_2_0.json
+  VALIDATOR_VERSION: "2.3.6"
+
+jobs:
+  interop:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      # On pull_request events this is the synthetic merge commit, i.e. the
+      # post-merge result.
+      - name: Check out PR head
+        uses: actions/checkout@v4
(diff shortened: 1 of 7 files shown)


halfcrazy:
report example https://github.com/halfcrazy/kubevirtbmc/actions/runs/33379984872/attempts/1#summary-99449915931

hack/redfish/interopserver/main.go
```diff
@@ -0,0 +1,102 @@
...
+	"syscall"
+
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	kubevirtv1 "kubevirt.io/api/core/v1"
+	cdifake "kubevirt.io/client-go/containerizeddataimporter/fake"
+	kvfake "kubevirt.io/client-go/kubevirt/fake"
+	crffake "sigs.k8s.io/controller-runtime/pkg/client/fake"
```
starbops: Just curious, what is this called `crffake`?

.github/workflows/redfish-interop.yml
starbops, line 1: Please use the latest versions of all referenced GitHub Actions.

.github/workflows/redfish-interop.yml
```diff
@@ -0,0 +1,109 @@
...
+        # infrastructure error, not a gate decision.
+        run: |
+          if [ -d logs/base ]; then
+            python3 hack/redfish/interop-diff.py --fail-on-introduced logs/head logs/base | tee -a "$GITHUB_STEP_SUMMARY"
+          else
+            python3 hack/redfish/interop-diff.py logs/head | tee -a "$GITHUB_STEP_SUMMARY"
+          fi
+
+      - name: Upload validator logs
+        if: always()
```
starbops, line 110: What's the reason for using `always()` here? I suppose if both of the preceding `run-interop.sh` steps fail, we don't need to trigger this step.

starbops (commented):
Thank you, this works great! Just few nits.

starbops (commented):
Also, it would be better to make it clear that the interoperability validation here is specifically for Ironic. We can extend this to other provisioning software in the future if they provide their profiles.

hack/redfish/interopserver/main.go
halfcrazy: emm, it was meant to cr(controller-runtime)fake, a duplicate f is a typo

.github/workflows/redfish-interop.yml
halfcrazy: If the Report step fails because the PR introduced new regressions (--fail-on-introduced), the full validator logs are the primary diagnostic material — the step summary only shows the diff, not the per-resource details.

.github/workflows/redfish-interop.yml
halfcrazy: done

.github/workflows/redfish-interop.yml
starbops: If that's the case, doesn’t the validator’s log already appear as the job log? Uploading the logs as an artifact seems redundant unless there’s important information that isn’t included in stdout.

halfcrazy: ConformanceLog_*.txt and InteropHtmlLog_*.html

starbops (approved):
LGTM, thank you!

anishbista60 (approved):
LGTM. Thanks for working on this

Issue #287, opened 2026-08-25
[TASK] Bump project dependencies to current stable versions


starbops:
Bump the project's dependencies to current stable versions for the upcoming 0.10.0 release, including:

- Go toolchain: 1.26
- Go modules (go.mod)
- Kubernetes client libraries (client-go, apimachinery, etc.): targeting 1.36
- KubeVirt and CDI API/client dependencies: targeting 1.9.0 and 1.66.0

anishbista60:
Any plan to bump the kind cluster version as well ?

starbops:
Kind itself is already at the latest version. The Kubernetes version used by the kind nodes has been bumped and aligned with the rest of the project.

anishbista60:
Make sure we shouldn't forget to update the  kubevirt version in docs as well.

PR #290, opened 2026-08-26
Dependency bump

starbops:
Besides the dependency bump, I also had to make a few additional changes to satisfy the new `golangci-lint` checks.

Related issue: #287


Makefile
```diff
@@ -122,11 +122,11 @@ test: manifests generate generate-kubevirt-crd fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $(shell go list ./... | grep -v /test/) -coverprofile cover.out
 
 GOLANGCI_LINT = $(shell pwd)/bin/golangci-lint
-GOLANGCI_LINT_VERSION ?= v2.6.1
+GOLANGCI_LINT_VERSION ?= v2.13.1
```
halfcrazy, line 125: Could we split out the golangci‑lint changes, or disable some of the new rules within this PR?
The modifications for duplicate string literals in the Go code changes seem overly strict.

Makefile
```diff
@@ -39,10 +39,10 @@ endif
 CONTAINER_TOOL ?= docker
 
 # KUBEVIRT API version to use (CRD codegen); keep in sync with KUBEVIRT_VERSION.
-KUBEVIRT_API_VERSION = v1.8.4
+KUBEVIRT_API_VERSION = v1.9.0
```
halfcrazy, line 42: We need to check the E2E results here. I'm concerned the KubeVirt version upgrade could introduce incompatible behavior for feature gates.

internal/controller/service/controller_test.go
```diff
@@ -38,6 +38,7 @@ var _ = Describe("Service Controller", func() {
 		testVMName                     = "test-vm"
 		testClusterIP                  = "10.0.0.100"
 		testSecretName                 = "test-secret"
+		testVirtualMachineBMCNoIPName  = "test-vmbmc-no-ip"
```
anishbista60: Is this flagged by linter ?

anishbista60 (commented):
Thanks for taking this up. I see there are changes to the code part, mostly renaming. Just wanna know, Is that all due to the lint warning ?

Makefile
starbops: The linting rules themselves haven't changed. It looks like the default rule set has simply become stricter.

I've split the actual dependency bump and the changes needed to satisfy the linting rules into two separate commits. We can loosen the rules a bit if needed. Let me know what you think.

Makefile
starbops: I haven't looked into it yet, but judging from the CI results and the tests running locally, it seems to be working unexpectedly well.

internal/controller/service/controller_test.go
starbops: Correct.

Makefile
anishbista60: I believe core kubevirt API will work as it is regardless of any nearest k8s version unless any breaking changes

Edited: I haven't seen last word "feature gates" lol :). Yes, we need to check that.

starbops:
> Thanks for taking this up. I see there are changes to the code part, mostly renaming. Just wanna know, Is that all due to the lint warning ?

Yes. The rule sets haven't changed a bit. It depends on our opinion whether to adjust the rules or fixing all of the warnings.

Makefile
starbops: That's a valid concern. If I recall correctly they've turned on all the beta feature gates since 1.9.0. There might be surprises.

Makefile
halfcrazy: Suggest ignoring this rule. `.golangci.yml` WDYT @anishbista60 
```
  linters:
    settings:
      goconst:
        ignore-map-keys: true
```

anishbista60: ok. I don't have much concern over here but since the PR is related to bump and yes we could also ignore those warning given by linter. Later, we can enable those linter one by one as part of cleanup process. 

Therfore, for now yes we could supress those lint issue for now as suggested by @halfcrazy

starbops: Sounds great. Let me adjust the rule set.

halfcrazy (dismissed):

starbops:
I've changed the default rule set to `standard`. Things became a lot easier.

halfcrazy (approved):
it looks much cleaner, thank you

.golangci.yml
```diff
@@ -3,33 +3,13 @@ run:
...
-    - unconvert
-    - unparam
-    - unused
-    # disabled: dupl, ginkgolinter, lll, revive, staticcheck
-  disable:
-    - dupl
-    - ginkgolinter
-    - lll
-    - revive
-    - staticcheck
```
anishbista60, line 26: rather then deleting it can't we simply disable ? 

oh okay you set `standard`

anishbista60 (approved):
LGTM. Thanks !

PR #297, opened 2026-09-07
Generate Redfish server stubs for declared operations

starbops:
Related issue: #302


pkg/redfish/emulator.go
```diff
@@ -24,13 +24,22 @@ type Emulator struct {
...
-func NewEmulator(ctx context.Context, port int, bmcUser string, bmcPassword string, resourceManager resourcemanager.ResourceManager) *Emulator {
+// newRouter builds the full Redfish route table: implemented routes only,
+// the session-creation response fixed up to match the Redfish spec, and
+// authentication required on everything but the public routes.
+func newRouter(bmcUser, bmcPassword string, resourceManager resourcemanager.ResourceManager) http.Handler {
 	apiService := NewAPIService(bmcUser, bmcPassword, resourceManager)
 	apiController := server.NewDefaultAPIController(apiService, server.WithDefaultAPIErrorHandler(recordingErrorHandler))
-	router := server.NewRouter(authFilter{
-		inner:      routeFilter{apiController},
+	return server.NewRouter(authFilter{
```
halfcrazy, line 33: openapi-generate provides the `addResponseHeaders` [option](https://github.com/OpenAPITools/openapi-generator/blob/master/docs/generators/go-server.md#config-options) for response headers, so we don't need a general middleware to filter on routes.

starbops: Thanks for the pointer! I'll check it out.

starbops: I'm afraid the middleware is still required, since post-processing, such as adding the allowable values, can only be done there. That said, the response header handling can be removed entirely.

halfcrazy: Maybe we can do it with a plan map
```
  func (s *APIService) RedfishV1SystemsComputerSystemIdGet(ctx context.Context, computerSystemId string)
  (server.ImplResponse, error) {
        computerSystem, err := s.handler.GetComputerSystem(ctx)
        if err != nil {
                return server.Response(http.StatusInternalServerError, nil), err
        }

        raw, err := json.Marshal(computerSystem)
        if err != nil {
                return server.Response(http.StatusInternalServerError, nil), err
        }
        var body map[string]any // magic comes from here
        if err := json.Unmarshal(raw, &body); err != nil {
                return server.Response(http.StatusInternalServerError, nil), err
        }
        if actions, ok := body["Actions"].(map[string]any); ok {
                if reset, ok := actions["#ComputerSystem.Reset"].(map[string]any); ok {
                        reset["ResetType@Redfish.AllowableValues"] = []server.ResourceResetType{
                                server.RESOURCERESETTYPE_ON,
                                server.RESOURCERESETTYPE_FORCE_OFF,
                                server.RESOURCERESETTYPE_GRACEFUL_SHUTDOWN,
                                server.RESOURCERESETTYPE_GRACEFUL_RESTART,
                                server.RESOURCERESETTYPE_FORCE_RESTART,
                        }
                }
        }

        return server.Response(200, body), nil
  }
```

starbops: Good suggestion. I think we can inline this for now and circle back to introducing middleware if we see more cases like this in the future.

halfcrazy: Yes. Currently, except for `auth/log` which exists as a generic middleware, everything else should be bypassable.

halfcrazy (dismissed):

halfcrazy (approved):

anishbista60 (approved):
LGTM. Thanks !

PR #298, opened 2026-09-08
feat: add standalone mode and publish static virtbmc binaries

### LINKED ISSUE #270: [FEATURE] Support running virtbmc without the operator

We ship KubeVirtBMC as an operator, and we assume everyone wants the operator. Not everyone does. vCluster's vMetal uses only `virtbmc` to demo their Metal3 flow against KubeVirt VMs, wiring it into their own manifests and their own lifecycle:

- https://vmetal.ai/docs/deploy/install#try-it-locally
- https://github.com/loft-sh/vcluster-bare-metal-with-kubevirt/blob/12796cd/vm/templates/bmc.yaml

They are on a fork of v0.7.0, and they cannot move forward. Since then `virtbmc` has grown a hard runtime dependency on being a Pod owned by a `VirtualMachineBMC` CR, so what was undocumented but possible is now impossible. Everything we have improved since v0.7.0 is unreachable to them.

This is worth caring about beyond one user. A BMC emulator is a building block. Requiring a CRD, a controller, and a webhook before anyone can emulate a single BMC is a steep entry price, and it pushes integrators toward forking instead of contributing.

Treat virtbmc as a component in its own right, with the operator as its most convenient consumer rather than its only entry point. Someone should be able to run one BMC emulator against one VM with nothing but an image, a ServiceAccount, and a Role.

That implies drawing a line: which parts of `virtbmc` are public contract, and which are ours to change freely. We have not drawn that line yet, and this issue is mostly about drawing it.

- Keep the operator as the only supported path. Integrators who need something smaller will keep forking, and we lose the feedback loop with them.
- Split `virtbmc` into its own repository. Solves the coupling by construction, but it splits the build, release, and e2e story for a single binary.

**Additional context**

Directions worth exploring, not decisions:

- Decouple BMC identity and state from the CR, so `virtbmc` can stand up on its own
- Firm up the CLI and credential contract, and document it
- Publish the standalone building blocks: RBAC, a reference manifest, maybe binaries
- Native TLS on the Redfish endpoint, so nobody needs a reverse-proxy sidecar

Regardless of this issue, we should reach out to the vCluster/vMetal folks about their fork and invite them to upstream the changes.

halfcrazy:
Allow running one BMC emulator against one VM with only an image, a ServiceAccount, and a Role — no VirtualMachineBMC CRD, controller, or cert-manager required.

Move the boot order controller to the agent, and it uses a local state file to persist status in standalone mode.

StateStore is a generic abstraction for persistent state. Switching the backend storage implementation is straightforward, and we can easily adopt standard ConfigMaps if needed.

fix https://github.com/kubevirtbmc/kubevirtbmc/issues/270


halfcrazy:
> Just one concern: we will have more configurations in the CRD in the long term. If we support all of them via virtbmc arguments, this can easily trigger an agent reload under managed mode. However, maintaining two separate approaches on the virtbmc side would introduce redundancy.

Addressed in a763ba04.

  The agent no longer takes the virtual media knobs as arguments: at InsertMedia time it GETs its own VirtualMachineBMC and
  reads the current spec, so a change takes effect on the next insert with no pod restart. Only listener/process-level settings (--enable-ipmi, ports, log level) and credentials stay restart-on-change.

  No second approach is introduced: there is one consumption path, and the per-mode config source mirrors the StateStore
  split (FileStateStore/ClusterStateStore) that standalone mode already established — in managed mode these flags simply
  don't exist, so there's no precedence logic either. Future CRD knobs of the same kind are just one more field read from
  the CR: no new flag, no controller rendering, no reload.

Issue #283, opened 2026-08-24
[BUG] DataVolume has no storageClassName, fails without a cluster-default StorageClass


patrostkowski:
**Describe the bug**

When a Redfish `VirtualMedia.InsertMedia` request is handled, `virtbmc` creates a `DataVolume` to back the inserted ISO (`ConstructDataVolume` in `pkg/util/helper.go`) without ever setting `spec.storage.storageClassName`. On a cluster with no default `StorageClass`, CDI cannot provision the backing PVC at all, so the `DataVolume` fails immediately and the inserted media never becomes available.

**To Reproduce**

1. Deploy a Kubernetes cluster with at least one `StorageClass` but **none** marked as default (`storageclass.kubernetes.io/is-default-class` absent on all of them).
2. Install KubeVirtBMC and create a `VirtualMachine` with an empty `cdrom` disk + a `VirtualMachineBMC` pointing at it.
3. Call `VirtualMedia.InsertMedia` against the BMC's Redfish endpoint with a valid ISO URL (or let an external caller like Metal3/Ironic's `redfish-virtualmedia` boot interface do it as part of inspection).
4. `kubectl describe datavolume <vm-name> -n <namespace>` shows:
```
   Reason:  ErrClaimNotValid
   Message: PVC spec is missing accessMode and no storageClass to choose profile
```
The media never actually attaches; the caller times out and retries the insert/eject cycle indefinitely.

**Expected behavior**

Either the `DataVolume` should be creatable against a specific, user-chosen `StorageClass` (no config surface exists for this today, checked `VirtualMachineBMCSpec`, the Helm chart values, and the controller's CLI flags, none expose it), or the failure should be surfaced clearly on the `VirtualMachineBMC`/Redfish response instead of silently retrying forever.

**Screenshots**

N/A

**Environment (please complete the following information):**

- OS: Ubuntu 24.04
- Kubernetes distribution/version: kubeadm v1.35.2
- Chart version: 0.6.0
- App version: v0.9.0

patrostkowski:
Fixed by https://github.com/kubevirtbmc/kubevirtbmc/pull/282

anishbista60:
Thank you @patrostkowski  for reporting this.  Let's include this  in `v0.10.0` release

/cc @starbops @halfcrazy

starbops:
[...]

PR #300, opened 2026-09-08
feat: pad DataVolume size via datavolume-size-margin annotation

### LINKED ISSUE #299: [BUG] Byte-exact DataVolume sizing breaks import on small volumes
**Describe the bug**

`VirtualMachineBMC`'s `InsertMedia` handler sizes the `DataVolume` it creates to the *exact* byte count reported by an HTTP `HEAD` request against the media URL (`GetRemoteFileSize` in `pkg/util/helper.go`), with zero margin:

```go
Requests: corev1.ResourceList{
    corev1.ResourceStorage: *resource.NewQuantity(size, resource.BinarySI),
},
```

CDI's HTTP-source import pipeline routes through a Filesystem-mode scratch PVC to buffer/convert the image regardless of the destination DataVolume's own volume mode. On a small (sub-GiB) volume, real filesystem metadata (superblock, inode tables, reserved blocks) consumes a disproportionate share of nominal capacity, so a byte-exact request fails even when the StorageClass's CDI filesystemOverhead is configured to 0 - that setting doesn't (and can't) make the real per-filesystem tax disappear.

I also tried working around it via VolumeMode: Block on the DataVolume (raw block device, no filesystem tax on the destination). Two related findings from that investigation:

1. Block mode needs AccessModes set explicitly on the DataVolume. CDI infers AccessModes from the StorageClass's StorageProfile when unset, and a profile that only declares a Filesystem claim-property-set (no Block entry - true of our Ceph RBD-backed StorageClass) has nothing to infer against, so CDI rejects the DataVolume outright: `Spec is missing accessMode and cannot get access mode from StorageProfile`
2. Block mode doesn't actually fix the root sizing problem described above - CDI still buffers through a Filesystem-mode scratch PVC internally, so the same failure recurs there instead. It also depends on the cluster having `device_ownership_from_security_context = true` set in containerd's CRI config, or the importer pod fails separately with `blockdev: cannot open /dev/cdi-block-volume: Permission denied` (a known, still-open CDI limitation - kubevirt/containerized-data-importer#2638).

To Reproduce

1. Configure a VirtualMachineBMC pointing at a VirtualMachine with a cdrom disk, storageClassName set to a Ceph RBD-backed (or any real block-device-backed) StorageClass.
2. Trigger VirtualMedia.InsertMedia via Redfish with a boot ISO a few hundred MB in size (e.g. an Ironic/IPA inspection or deploy ramdisk ISO).
3. Watch the resulting DataVolume's importer pod logs.

Expected behavior

The DataVolume import succeeds - the requested size should include enough margin to survive real filesystem overhead on the storage backend, not just the exact reported Content-Length.

Screenshots

N/A - see importer pod log excerpt below instead:

```
E0907 11:19:06.210012 1 data-processor.go:256] virtual image size 333688832 is larger than the reported available storage 297829376. A larger PVC is required
Unable to transfer source data to scratch space
```

(Requested/reported Target size: 334495744; real usable space on the scratch PVC: 297829376 - an ~11% shortfall on a ~318MiB image.)

Environment (please complete the following information):

- OS: Ubuntu 24.04.2 LTS
- Kubernetes distribution/version: kubeadm v1.35.8
- Chart version: 0.7.0
- App version: v0.10.0
- Storage: Rook-Ceph RBD CSI (rook-ceph.rbd.csi.ceph.com), CDI filesystemOverhead for the StorageClass explicitly set to 0
- Container runtime: containerd 2.2.1

**Additional context**

I worked around this locally with a fixed 30% margin on the requested `DataVolume` size, but I know #131/#170 already covered this ground, and a fixed margin PR was closed.

That workaround doesn't help us specifically: our cluster pins `filesystemOverhead` to `0` for this StorageClass as platform policy (avoids Ceph RBD's coarse rounding doubling unrelated tenant volumes), so I can't tune it per-consumer.

Would an opt-in annotation be an acceptable middle ground instead of a `spec` field e.g. `bmc.kubevirt.io/datavolume-size-margin: "30"` on the `VirtualMachineBMC`, defaulting to `0` when absent? I already have a PR ready implementing this and will open it against this issue once agreed upon the solution.

patrostkowski:
Fixes https://github.com/kubevirtbmc/kubevirtbmc/issues/299


pkg/resourcemanager/virtualmachine_resourcemanager.go
```diff
@@ -221,22 +222,40 @@ func (m *VirtualMachineResourceManager) InsertMedia(ctx context.Context, imageUR
...
 			}
-		} else if bmc.Spec.StorageClassName != nil {
-			storageClassName = *bmc.Spec.StorageClassName
+		} else {
+			if bmc.Spec.StorageClassName != nil {
+				storageClassName = *bmc.Spec.StorageClassName
+			}
+			volumeMode = bmc.Spec.VolumeMode
+			if parsed, err := strconv.Atoi(bmc.Annotations[bmcv1.AnnotationDataVolumeSizeMargin]); err == nil {
+				sizeMarginPercent = parsed
```
halfcrazy: Invalid margins fail silently

pkg/util/helper.go
```diff
@@ -53,32 +61,47 @@ func GetRemoteFileSize(url string) (int64, error) {
...
+	// StorageClassName falls back to the cluster default when empty.
+	StorageClassName string
+	// VolumeMode falls back to CDI's own default (Filesystem) when nil.
+	VolumeMode *corev1.PersistentVolumeMode
+}
+
+// ConstructDataVolume builds the DataVolume backing an inserted virtual media image.
+func ConstructDataVolume(params DataVolumeParams) *cdiv1.DataVolume {
 	storage := &cdiv1.StorageSpec{
+		// Explicit, since a StorageClass with no Block claim property set can't be inferred against for VolumeMode: Block.
```
halfcrazy:
```
storage := &cdiv1.StorageSpec{
    VolumeMode: params.VolumeMode,
    Resources:  ...,
}
if params.VolumeMode != nil {
    storage.AccessModes = []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce}
}
```

api/bmc/v1beta1/virtualmachinebmc_types.go
```diff
@@ -52,6 +55,14 @@ type VirtualMachineBMCSpec struct {
...
 	StorageClassName *string `json:"storageClassName,omitempty"`
+
+	// VolumeMode is the volume mode for the DataVolume created on virtual media insert; unset keeps
+	// today's behavior (Filesystem, CDI's own default). Block requests a raw block device instead,
+	// which has no filesystem overhead and so isn't subject to the StorageClass's CDI
+	// filesystemOverhead setting — useful when that setting can't accommodate an exact-size image.
+	// +optional
+	// +kubebuilder:validation:Enum=Filesystem;Block
+	VolumeMode *corev1.PersistentVolumeMode `json:"volumeMode,omitempty"`
 }
```
anishbista60, line 96: Since we recently added the `storageClassName` API, we should have discussed it more deeply earlier, as mentioned in this comment: https://github.com/kubevirtbmc/kubevirtbmc/issues/283#issuecomment-5397601466 but better late than never. 

I'm thinking of moving that API under `DataVolumeSpec`, which would help keep the root API clean and all configuration related to the DataVolume should go under `DataVolumeSpec`.

```yaml
type DataVolumeSpec struct {
    StorageClassName  *string                       `json:"storageClassName,omitempty"`
    VolumeMode        *corev1.PersistentVolumeMode  `json:"volumeMode,omitempty"`
}
```

```yaml
// on VirtualMachineBMCSpec:
DataVolume *DataVolumeSpec `json:"dataVolume,omitempty"`
```


Having said that, this will introduce the breaking changes for these api in next release and we should mentioned this clearly in the release note.

anishbista60: @halfcrazy @starbops WDYT ?

api/bmc/v1beta1/virtualmachinebmc_types.go
```diff
@@ -31,6 +31,9 @@ const (
 	VMNameLabel                      = "kubevirt.io/vm-name"
 )
 
+// AnnotationDataVolumeSizeMargin pads the inserted-media DataVolume's size by this many percent; absent/invalid defaults to 0.
+const AnnotationDataVolumeSizeMargin = "bmc.kubevirt.io/datavolume-size-margin"
```
anishbista60: Please move this constant above

api/bmc/v1beta1/virtualmachinebmc_types.go
starbops: Yes, we should group them under their respective categories. At the same time, I’d prefer to keep the intention clear while exposing as little of the implementation details to users as possible. From their perspective, they should only need to interact with the Redfish/IPMI interfaces without having to understand how those capabilities are implemented underneath.

That said, I can also see the counterargument that people creating VirtualMachineBMC CRs are typically administrators or infrastructure operators, so it may be reasonable to expect them to have more knowledge of the underlying implementation than regular end users. I agree with that to some extent as well, so I think this is still open for discussion.

My original thought was to group them under the `spec.redfish` field. For instance:

```yaml
spec:
  redfish:
    storageClassName: fast
    volumeMode: Block
```

Or, to be more specific, imagine that we’ll have more configurable options in the future and want to keep them well structured:

```yaml
spec:
  redfish:
    virtualMedia:
      storageClassName: fast
      volumeMode: Block
```

Does the above make sense to you?

anishbista60: I share the concern about `virtualMedia` as a group name. Both fields are consumed by the DataVolume, but naming the API after the implementation is what we're trying to avoid, and it locks us in if we ever change how the media gets materialized.

I'd suggest storage instead:

``` yaml
spec:
  redfish:
    virtualMedia:
      storage:
        storageClassName: fast
        volumeMode: Block
   ```

It mirrors `DataVolume.spec.storage` (and `PersistentVolumeClaimSpec`) without naming CDI, so the controller can copy the struct straight across, and there's room for accessModes or a size override later.

Comment (anishbista60) [reply api/bmc/v1beta1/virtualmachinebmc_types.go:96]: Ref: https://github.com/kubevirt/containerized-data-importer/blob/main/staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go#L74

Comment (anishbista60) [reply api/bmc/v1beta1/virtualmachinebmc_types.go:96]: I have one  more question, Does IPMI support media mount ? If yes, what is the plan for addition more functionality for that one ? 

/cc @halfcrazy You might have more idea on this cause you are working on ipmi library at upstream.

Comment (halfcrazy) [reply api/bmc/v1beta1/virtualmachinebmc_types.go:96]: > I have one more question, Does IPMI support media mount ? If yes, what is the plan for addition more functionality for that one ?
> 
> /cc @halfcrazy You might have more idea on this cause you are working on ipmi library at upstream.

AFAIK, there is no equivalent virtualmedia in the IPMI spec.

Comment (starbops) [reply api/bmc/v1beta1/virtualmachinebmc_types.go:96]: That falls into OEM commands territory, and we currently have no plans to support it.

Comment (starbops) [reply api/bmc/v1beta1/virtualmachinebmc_types.go:96]: To avoid losing focus, @patrostkowski, would you mind slightly adjusting the `volumeMode` field and placing it where @anishbista60 suggested?

```yaml
spec:
  redfish:
    virtualMedia:
      storage:
        storageClassName: fast
        volumeMode: Block
```

We can handle the `storageClassName` migration separately in a follow-up task. WDYT?

anishbista60: yep. We can handle migration in next followup PR. 

Thank you all for your valuable insights.

pkg/resourcemanager/virtualmachine_resourcemanager.go
```diff
@@ -221,22 +222,40 @@ func (m *VirtualMachineResourceManager) InsertMedia(ctx context.Context, imageUR
...
+			}
 		}
 	}
 
+	imageSize = util.WithImportMargin(imageSize, sizeMarginPercent)
+
 	// Create DataVolume
-	dv := util.ConstructDataVolume(m.namespace, m.name, imageURL, imageSize, storageClassName)
+	dv := util.ConstructDataVolume(util.DataVolumeParams{
```
anishbista60: Nit: 
```suggestion
	dv := util.ConstructDataVolume(util.DataVolumeOptions{
```

halfcrazy (dismissed):

pkg/resourcemanager/virtualmachine_resourcemanager.go
```diff
@@ -221,22 +222,45 @@ func (m *VirtualMachineResourceManager) InsertMedia(ctx context.Context, imageUR
 		return err
 	}
 
-	// A missing BMC client/object means no StorageClassName override is configured, not a failure.
+	// A missing BMC client/object means no StorageClassName/VolumeMode/size-margin override is
+	// configured, not a failure.
 	var storageClassName string
+	var volumeMode *corev1.PersistentVolumeMode
```
anishbista60:
```suggestion
var (
	storageClassName string
	volumeMode       *corev1.PersistentVolumeMode
)
```

pkg/resourcemanager/virtualmachine_resourcemanager.go
```diff
@@ -221,22 +222,45 @@ func (m *VirtualMachineResourceManager) InsertMedia(ctx context.Context, imageUR
...
+	sizeMarginPercent := 0
```
anishbista60:
```suggestion
	sizeMarginPercent int
```

move this in above var cause by default it would be `0`

halfcrazy (approved):

anishbista60 (approved):
LGTM. Thanks for working on it.

starbops (approved):
LGTM. Nice improvement. Thanks for working through the details!

PR #305, opened 2026-09-10
test(e2e): tolerate Node Busy on repeated power-off during teardown

### LINKED ISSUE #304: [BUG] Agent e2e: repeated power-off returns Node Busy when the VMI teardown outruns the second call
**Describe the bug**

`Agent e2e › IPMI operations › Power management › should treat repeated power-off as idempotent`
intermittently fails: the second `chassis power off` returns 0xC0 **Node Busy** instead of success.

`PowerOff` calls KubeVirt's `Stop` and, if that errors, re-reads the VM — reporting success only once
`vm.Status.Ready` is false. Right after the first `power off`, KubeVirt has not necessarily flipped `Ready`
yet, so the re-read still sees the VM as running and the handler returns the retryable error that
`pkg/resourcemanager/errors.go` maps to Node Busy ("command processing resources are temporarily
unavailable", IPMI 2.0 §5.2 — the code clients are expected to retry). The flake is in the spec assuming
that flip always beats the second call.

Observed in CI: https://github.com/kubevirtbmc/kubevirtbmc/actions/runs/34438778626/job/102749256559
(`[FAILED] [0.420 seconds]`, `stderr="Set Chassis Power Control to Down/Off failed: Node busy\n"`;
the other 21 specs that ran before the ordered suite stopped passed).

**To Reproduce**

Timing-dependent, not reproducible on every run:

1. `make local-e2e-test` (kind + KubeVirt), or deploy the e2e env and run
   `go test -v -timeout 15m ./test/virtbmc-agent/...`
2. Re-run the agent suite; the failure shows up when the second `power off` lands before KubeVirt flips
   `vm.Status.Ready`.

**Expected behavior**

The second `power off` should eventually succeed: retry while the handler answers Node Busy — the shape the
power-on race spec (`should return retryable error when power-on races with VMI cleanup`) already uses — and
keep asserting the VM ends up stopped (`waitForVMIDeleted`), so the "power actions are not swallowed"
protection from #220 stays in place.

**Screenshots**

Not applicable (CI log linked above).

- OS: GitHub Actions runner, job `build-pr`
- Kubernetes distribution/version: kind v1.36.1
- Chart version: n/a (e2e deploys the manifests directly)
- App version: build of `c1c0faba`

**Additional context**

- The strict assertion comes from #220 ("retry instead of swallowing power actions during VM transitional
  states"); this is the same class of race, just in the power-off direction.
- Fix drafted on branch `flake-repeated-power-off` (d8b50c73): wrap the second call in `Eventually`,
  tolerating only Node Busy.

halfcrazy:
Fixes #304.

The second `chassis power off` in this spec can land before KubeVirt flips
`vm.Status.Ready`, and the handler then answers Node Busy (0xC0, the retryable
completion code) instead of success — a matter of the controller's reconcile
timing, not a regression, which is what made the spec flaky.

Retry the way the power-on race spec already does, tolerating only Node Busy, and
keep the `waitForVMIDeleted` assertion so the "power actions are not swallowed"
protection from #220 stays in place.

Test-only change, no production code touched. The flake is only observable in the
kind e2e, so CI is the confirmation.


anishbista60 (approved):
LGTM. Thanks for fix !

starbops (approved):
Thanks for fixing the flaky test. LGTM.

PR #310, opened 2026-09-11
Make InsertMedia idempotent for same-image retries

rsacherer:
## Summary

This change makes `InsertMedia` safe to call more than once for the same image URL.

Ironic ejects media before every insert, but the eject is best-effort. When the eject fails (pod restart, network timeout), the next `InsertMedia` call finds the DataVolume and volume attachment still present from the previous call. The current code fails with `AlreadyExists` on the DataVolume creation and duplicates the volume entry in the VM spec.

## Behavior after this change

- Same image URL as the existing DataVolume: treat the call as a retry and succeed.
- Different image URL: return an error. The caller must eject first.
- Volume already attached in the VM spec: skip the attachment and the VM update.

This preserves the original failure semantics for conflicting media while making retries of the same image safe.

## Changes

- `pkg/resourcemanager/virtualmachine_resourcemanager.go`:
  - On `AlreadyExists`, fetch the existing DataVolume and compare its source URL. Succeed if the URL matches. Fail with a descriptive error if it does not.
  - Check if the volume is already in the VM spec before appending.
- `pkg/resourcemanager/virtualmachine_resourcemanager_test.go`:
  - Rename the existing "should fail" test case to cover the different-image conflict case.
  - Add a new test case for the same-image idempotent retry.

## Why fixing it in kubevirtbmc

kubevirtbmc returns a generic HTTP 500 for any `InsertMedia` error. Ironic cannot distinguish "already inserted" from a real failure. The Redfish spec defines no standard error code for this case. Fixing it at the kubevirtbmc layer avoids a two-repo change with no spec to lean on.


pkg/resourcemanager/virtualmachine_resourcemanager.go
```diff
@@ -284,25 +284,46 @@ func (m *VirtualMachineResourceManager) InsertMedia(ctx context.Context, imageUR
...
-	volume := kubevirtv1.Volume{
-		Name: cdromDisk.Name,
-		VolumeSource: kubevirtv1.VolumeSource{
-			DataVolume: &kubevirtv1.DataVolumeSource{
-				Name:         dv.Name,
-				Hotpluggable: true,
-			},
-		},
+	// Attach DataVolume to VirtualMachine (skip if already attached with the same source)
+	volumeExists := false
```
halfcrazy:
```
volumeExists := false
for _, v := range vm.Spec.Template.Spec.Volumes {
    if v.Name != cdromDisk.Name {
        continue
    }
    if v.VolumeSource.DataVolume == nil || v.VolumeSource.DataVolume.Name != dv.Name {
        return fmt.Errorf("cdrom volume %q already attached to a different source", cdromDisk.Name)
    }
    volumeExists = true
    break
}
```

anishbista60 (commented):
@rsacherer Thanks for working on it. Please add the e2e test for it as well.

PR #312, opened 2026-09-12
fix(redfish): implement SessionCollection GET to stop 404/405 flip

### LINKED ISSUE #311: [BUG] Redfish: GET SessionService/Sessions flakes between 404 and 405
**Describe the bug**

Since #269 stripped unimplemented Redfish routes, `GET /redfish/v1/SessionService/Sessions` nondeterministically returns **404 or 405**. The status is decided once per process start by the random order in which gorilla/mux registers routes (map iteration order of the generated route table).

The Redfish-Interop-Validator CI job (#288) diffs base vs. head runs by exact failure message, so two runs of *identical code* landing on different statuses surfaces as a phantom "introduced" failure on unrelated PRs — e.g. https://github.com/kubevirtbmc/kubevirtbmc/actions/runs/34670917882 failed on #298, which touched no routing code at all.

**To Reproduce**

1. Start the Redfish emulator (or `hack/redfish/interopserver`).
2. `curl -u admin:password http://localhost:8000/redfish/v1/SessionService/Sessions`
3. Restart the process and repeat.
4. The status flips: ~2/3 of the runs answer 404, ~1/3 answer 405.

Route-table-only repro: build the router N times in a test and tally the status — 50 builds gave 29×404 / 21×405.

**Expected behavior**

A stable status code. Per the Redfish spec the Sessions collection exists (POST creates sessions there), so GET should be implemented and answer 200 with a valid `SessionCollection`.

**Screenshots**

N/A

- OS: N/A (reproduces in CI)
- Kubernetes distribution/version: N/A
- Chart version: N/A
- App version: main @ 981059b

**Additional context**

Root cause:

- `routeFilter` (#269) deletes `RedfishV1SessionServiceSessionsGet` (a 501 stub) but keeps `RedfishV1SessionServiceSessionsPost` on the same path.
- In gorilla/mux, the POST route records `ErrMethodMismatch` for a GET request, but any *later-registered* GET route whose method matcher succeeds clears `MatchErr` again (`route.go`), downgrading the outcome to 404. Only when the POST route happens to register last among the unprotected routes does 405 survive.
- `hack/redfish/interop-diff.py` compares validator messages verbatim, so `(404)` vs `(405)` becomes an introduced+fixed pair that fails the regression gate.

While at it, the session lifecycle has three pre-existing inconsistencies that a real collection would expose: `DELETE /SessionService/Sessions/{id}` is a no-op (passes the session ID to `RemoveToken`, which keys on the token), the SessionsPost body hardcodes `@odata.id` to `.../Sessions/1` while the `Location` header carries the real ID, and Session payloads miss the `#` prefix in `@odata.type`.

Fix in progress; PR to follow.

halfcrazy:
fix https://github.com/kubevirtbmc/kubevirtbmc/issues/311

## Root cause

Since #269 stripped unimplemented routes, `RedfishV1SessionServiceSessionsGet` (a 501 stub) has no route while `SessionsPost` on the same path stays registered. gorilla/mux registers routes by iterating a map, so the outcome for GET on that path is decided per process start:

- the POST route records `ErrMethodMismatch` for the GET request;
- but any *later-registered* GET route whose method matcher succeeds clears `MatchErr` again (`route.go`), downgrading to 404.

Only when the POST route lands last among the unprotected routes does 405 survive (~1/3 of process starts; measured 21×405 / 29×404 over 50 router builds). The interop CI diffs base vs. head by exact validator message, so two runs of identical code rolling different statuses shows up as a phantom "introduced" failure — e.g. https://github.com/kubevirtbmc/kubevirtbmc/actions/runs/34670917882 on #298.

## Fix

Implement `RedfishV1SessionServiceSessionsGet` for real: list the live sessions from the token store. The endpoint now deterministically answers 200 with a valid `SessionCollection`, so the validator failure disappears from both base and head runs and the gate settles. (Making the stripped stubs answer 501 was rejected: it would still change the message vs. base and fail the gate on this very PR; normalizing status codes in interop-diff would only hide the server-side nondeterminism.)

Building the collection surfaced three pre-existing inconsistencies it would otherwise expose, fixed in the same commit:

- `DELETE /SessionService/Sessions/{id}` was a no-op — it passed the session ID to `RemoveToken`, which keys on the token. Terminated sessions kept authenticating and would have stayed listed. Now removed by session ID.
- `SessionsPost` hardcoded the body `@odata.id` to `.../Sessions/1` (never resolvable) while the `Location` header carried the real ID — aligned with the header.
- Session payloads used `Session.v1_7_1.Session` without the `#` the OData type URL requires (DSP0266) — every other resource has it.

Also fixed `TestDeleteSession`, which asserted via `GetToken(sessionID)` and therefore passed vacuously.

## Testing

- `TestGetSessionCollection` + fixed `TestDeleteSession`; `go build/vet/test ./pkg/...` green.
- Determinism: 50 router builds → 50×401 (no auth) / 50×200 (basic auth); previously 29×404 / 21×405.
- End-to-end against `hack/redfish/interopserver`: POST → collection lists the member with the canonical URI; member GET 200; DELETE → 204; afterwards member GET 404, collection empty, old token 401.
- The validator runs with `authtype=Basic, forceauth=True` and never creates sessions, so it sees a stable empty collection and the `could not be acquired` failure is simply gone.

### DIFF
--- a/pkg/redfish/api_service.go
+++ b/pkg/redfish/api_service.go
@@ -53816,16 +53816,7 @@ func (s *APIService) RedfishV1SessionServicePatch(ctx context.Context, sessionSe
 
 // RedfishV1SessionServiceSessionsGet -
 func (s *APIService) RedfishV1SessionServiceSessionsGet(ctx context.Context) (server.ImplResponse, error) {
-	// TODO - update RedfishV1SessionServiceSessionsGet with the required logic for this service method.
-	// Add api_default_service.go to the .openapi-generator-ignore to avoid overwriting this service implementation when updating open api generation.
-
-	// TODO: Uncomment the next line to return response Response(200, SessionCollectionSessionCollection{}) or use other options such as http.Ok ...
-	// return Response(200, SessionCollectionSessionCollection{}), nil
-
-	// TODO: Uncomment the next line to return response Response(0, RedfishError{}) or use other options such as http.Ok ...
-	// return Response(0, RedfishError{}), nil
-
-	return server.Response(http.StatusNotImplemented, nil), errors.New("RedfishV1SessionServiceSessionsGet method not implemented")
+	return server.Response(200, s.handler.GetSessionCollection()), nil
 }
 
 // RedfishV1SessionServiceSessionsPost -
@@ -53848,8 +53839,8 @@ func (s *APIService) RedfishV1SessionServiceSessionsPost(ctx context.Context, se
 	}
 
 	return server.Response(201, server.SessionV171Session{
-		OdataType: "Session.v1_7_1.Session",
-		OdataId:   "/redfish/v1/SessionService/Sessions/1",
+		OdataType: "#Session.v1_7_1.Session",
+		OdataId:   fmt.Sprintf("/redfish/v1/SessionService/Sessions/%s", id),
 		Id:        id,
 		Name:      "User Session",
 		Token:     &token, // The final response will not contain this field. It's just a means to pass the token to the caller.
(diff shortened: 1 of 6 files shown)


starbops (approved):
LGTM, thanks for the fix!

anishbista60 (approved):

PR #316, opened 2026-09-15
fix(redfish): tolerate empty body in parameterless action POSTs

### LINKED ISSUE #315: [BUG] Redfish EjectMedia rejects empty request body, breaking Metal3/Ironic deploys (regression from #297)
**Describe the bug**

Since #297 (`aeacfb1`, "build(redfish): keep generated API limited and reproducible"), the Metal3 e2e job fails on `main`: the Ironic node goes to `deploy failed` at the `deploy.prepare_instance_boot` step while ejecting virtual media.

Root cause: the regenerated `pkg/generated/redfish/server/api_default.go` no longer tolerates an **empty request body** on `POST /redfish/v1/Managers/{ManagerId}/VirtualMedia/{VirtualMediaId}/Actions/VirtualMedia.EjectMedia`. sushy (used by Ironic/Metal3) sends this action with **no body and no Content-Type**. The generated controller unconditionally decodes the body, gets `io.EOF`, and returns `400 ParsingError`. Worse, the generated `DefaultErrorHandler` encodes the error as a bare JSON string (`"EOF"`) instead of a Redfish `{"error": {...}}` object, so sushy crashes inside its own error handling (`body.get('error', {})` on a `str`) and the real error is masked.

This is a **regression of the bug fixed in #180** (`83ea326`, "Fix: tolerate empty body in EjectMedia handler for VirtualMedia"). That fix was a manual edit to the generated `api_default.go`; #297 made generation fully reproducible (regenerating `api_default.go` every run, with `hack/redfish/relax-patch-assertions` re-applying the PATCH-assertion relaxation), but the EjectMedia empty-body tolerance had no such post-generation re-apply step and was silently overwritten.

Failing run: https://github.com/kubevirtbmc/kubevirtbmc/actions/runs/34940412329

**To Reproduce**

1. Deploy kubevirtbmc at `aeacfb1` (or any build with the regenerated redfish server).
2. Register a BMH against its Redfish endpoint and start provisioning (or simply: `curl -X POST http://<bmc>/redfish/v1/Managers/BMC/VirtualMedia/CD1/Actions/VirtualMedia.EjectMedia` with an empty body).
3. The request returns `400` with a plain-string body instead of `204`.

**Expected behavior**

`VirtualMedia.EjectMedia` with an empty body is accepted (treated as "no parameters supplied") and returns `204 No Content`, as it did before #297 and as fixed in #180. Per the Redfish spec clients should send `{}`, but tolerating a completely empty body is required for compatibility with sushy/Ironic.

**Screenshots / Logs**

From the Metal3 e2e diagnostics artifact (`ironic-ironic.log`), node `f2d3bbf9-5de0-4cc6-b452-d08a883e87b3`:

```
2026-09-15 07:24:59.412 62 DEBUG sushy.connector [...] HTTP request:
  POST http://10.96.223.172/redfish/v1/Managers/BMC/VirtualMedia/CD1/Actions/VirtualMedia.EjectMedia;
  headers: {'OData-Version': '4.0', 'Accept-Encoding': 'identity'}; body: None

2026-09-15 07:24:59.413 62 ERROR ironic.drivers.modules.agent [...]
  Preparing instance for booting failed for instance 1c1185fc-.... AttributeError: 'str' object has no attribute 'get'
  ...
  File ".../sushy/resources/manager/virtual_media.py", line 300, in eject_media
    self._conn.post(target_uri)
  File ".../sushy/exceptions.py", line 179, in raise_for_response
    raise BadRequestError(method, url, response)
  File ".../sushy/exceptions.py", line 113, in __init__
    self.body = body.get('error', {})
  AttributeError: 'str' object has no attribute 'get'

2026-09-15 07:25:00.247 62 ERROR ironic.conductor.task_manager [...]
  Node f2d3bbf9-... moved to provision state "deploy failed" from state "deploying"
```

The same crash repeats during node tear-down (`Error in tear_down of node f2d3bbf9-...: AttributeError: 'str' object has no attribute 'get'`).

- App version: `main` @ `aeacfb1` (kubevirtbmc image built by CI)
- Client: sushy (Ironic/Metal3 baremetal-operator e2e, kind cluster)

**Additional context**

Only `EjectMedia` is affected: it is the one action sushy invokes with an empty body (`ComputerSystem.Reset` and `VirtualMedia.InsertMedia` carry JSON bodies).

Suggested fix, following the pattern #297 already established:

1. Add a post-generation step to `hack/redfish/generate.sh` (alongside `relax-patch-assertions`) that re-applies the empty-body (`io.EOF`) tolerance to the EjectMedia controller in `api_default.go`, restoring the behavior from #180 in a reproducible way.
2. Optionally, inject a custom `ErrorHandler` that renders parsing errors as a proper Redfish `{"error": {...}}` payload instead of a bare JSON string, so future 400s are diagnosable from sushy/Ironic logs instead of surfacing as `AttributeError: 'str' object has no attribute 'get'`.

Unrelated but noticed in the same logs (non-fatal, Ironic tolerates them): `The attribute Bios is missing from the resource /redfish/v1/Systems/1` and `The attribute Links/Chassis is missing from the resource /redfish/v1/Systems/1`. May be worth a separate check whether the trimmed generated API dropped fields the previous implementation exposed.

halfcrazy:
assertion relaxation via relax-patch-assertions, but the empty-body tolerance for EjectMedia from #180 (a hand-edit to the generated file) had no such post-generation step and was silently overwritten. sushy (Ironic/Metal3) sends VirtualMedia.EjectMedia with no body at all, so the decode guard's io.EOF became a 400 and Metal3 e2e deploys failed at prepare_instance_boot.

Add tolerate-empty-action-body, a post-generation fixup in the same mold as relax-patch-assertions, and wire it into generate.sh. It relaxes the decode guard only for handlers whose request body is a free-form map[string]interface{} (EjectMedia, SetDefaultBootOrder); typed-body handlers (InsertMedia, Reset) keep strict decoding.

Fixes #315


test/metal3-e2e/trigger
starbops, line 1: Do you know what this file is used for?

starbops (commented):
Thanks for the fix. Just one nit.

test/metal3-e2e/trigger
halfcrazy: I need to trigger CI before merging into main. Since this E2E test is heavy, it won’t be automatically triggered for every PR.
https://github.com/kubevirtbmc/kubevirtbmc/blob/aeacfb1335cae2f0a56b36c00c52511662969616/.github/workflows/metal3-e2e.yml#L11-L15

starbops: Ah, I see. I appreciate the reminder.

starbops (approved):
LGTM, thank you.

anishbista60 (approved):
Thank you for fix 😍

Issue #317, opened 2026-09-16
[ENHANCEMENT] Trigger metal3-e2e by label on demand


halfcrazy:
**Is your enhancement request of an existing feature related to a problem? Please describe.**

Trigger metal3-e2e by label on demand even if no harness-related files have been modified

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Add any other context or screenshots about the enhancement request here.

PR #318, opened 2026-09-16
ci(metal3-e2e): allow label-triggered runs on any PR

halfcrazy:
The Metal3 e2e suite only ran on PRs touching its harness paths, so a change to Redfish/IPMI behavior could not be verified against it before merge; PRs resorted to bumping a dummy timestamp file under test/metal3-e2e/ to fool the paths filter.

Make the workflow reusable (workflow_call) and add a thin wrapper workflow that invokes it when the 'need-metal3-e2e' label is applied. Applying labels requires triage permission or above, so the label itself is the authorization.

Also add a concurrency group: repeated triggers on the same PR cancel stale in-progress runs instead of letting 40-70 min jobs pile up, while branch/tag runs keep queueing so each merged commit retains its own test record.

Drop the now-obsolete dummy trigger file and document the label in CONTRIBUTING.md.

#317


halfcrazy:
https://github.com/halfcrazy/kubevirtbmc/pull/3 test: trigger metal3 e2e by path
https://github.com/halfcrazy/kubevirtbmc/pull/4 test: trigger metal3 e2e by label
https://github.com/halfcrazy/kubevirtbmc/pull/6 test: do not trigger metal3 e2e

halfcrazy:
@dosu Reviewing the workflow/concurrency-group logic for correctness

halfcrazy:
@dosu Reviewing the workflow/concurrency-group logic for correctness

anishbista60 (approved):
LGTM. Thanks man for the quick fix ❤️

starbops (approved):
LGTM, thank you!

PR #319, opened 2026-09-17
ci(redfish): add Redfish-Service-Validator workflow

halfcrazy:
The interop workflow checks Ironic profile conformance (presence, link reachability, required values), but never validates payloads against the schemas: a resource that answers 200 with the right @odata.type passes no matter how broken the body is. The DMTF Redfish-Service-Validator fills that gap, checking every reachable resource against the CSDL schemas from the local DSP8010 bundle -- the same schema version the generated API is built from, so no profile file and no live dmtf.org access is needed.

Mirrors the interop tooling: run-service-validate.sh (local + CI entry, make redfish-service-validate), service-diff.py (head/base diff with --fail-on-introduced), and a workflow that gates PRs on introduced failures only, so the existing schema violations on main don't block unrelated PRs.

sister pr of #288 for #275


halfcrazy:
https://github.com/halfcrazy/kubevirtbmc/pull/7

starbops (approved):
LGTM. Really appreciate the PR! I believe this completes the picture, as we now have general E2E testing, Metal3 E2E testing, Redfish spec conformance checks, and the Ironic interoperability validator. Together, they not only help catch regressions but also send a clear message that we’re committed to playing nicely with the Redfish spec and the broader ecosystem.

anishbista60 (approved):
LGTM. Thanks ! 

Feel free to merge it.

